PR #2, opened 2025-03-21
Add S-CORE CI jobs for baselib repository

nradakovic:
The current list of jobs si not definite and more will be added once the tools that are executed in workflows are added.


antonkri (approved):

.github/workflows/format.yml
```diff
@@ -0,0 +1,30 @@
...
+# information regarding copyright ownership.
+#
+# This program and the accompanying materials are made available under the
+# terms of the Apache License Version 2.0 which is available at
+# https://www.apache.org/licenses/LICENSE-2.0
+#
+# SPDX-License-Identifier: Apache-2.0
+# *******************************************************************************
+
+name: Formatting checks
```
LittleHuba, line 14: What is the naming convention here? Other jobs have `S-CORE CI -` prepended. This one does not. Is that intentional?

.github/workflows/gitlint/action.yml
```diff
@@ -0,0 +1,42 @@
...
+# information regarding copyright ownership.
+#
+# This program and the accompanying materials are made available under the
+# terms of the Apache License Version 2.0 which is available at
+# https://www.apache.org/licenses/LICENSE-2.0
+#
+# SPDX-License-Identifier: Apache-2.0
+# *******************************************************************************
+
+name: "Gitlint Action"
```
LittleHuba, line 14: What is the naming convention here? Other jobs have S-CORE CI - prepended. This one does not. Is that intentional?

BUILD
```diff
@@ -0,0 +1,48 @@
...
+# This program and the accompanying materials are made available under the
+# terms of the Apache License Version 2.0 which is available at
+# https://www.apache.org/licenses/LICENSE-2.0
+#
+# SPDX-License-Identifier: Apache-2.0
+# *******************************************************************************
+load("@aspect_rules_lint//format:defs.bzl", "format_multirun", "format_test")
+load("@score_cr_checker//:cr_checker.bzl", "copyright_checker")
+
+# TODO: Add all files/directories in `src` attribute that needs to be checked
```
LittleHuba, line 16: Does this apply to all rules below or just to this one?

nradakovic: I can add it inside argument for better understanding, but again other rules do not have `src` argument.

LittleHuba: That is good enough for me. Just wanted to clarify.

nradakovic:
The PR is outdate, new version will be added over mirror.

PR #3, opened 2025-05-15
fix: compile issues, types missing

MarkSchmitt:
When trying to build communication, my compiler was complaining about some missing symbols that seem to be caused by missing includes.


ramceb:
@MarkSchmitt thanks for the report. We are working on it. Your fix is valid but does not pass some minor clang-format checks, which are not yet established in score.

PR #11, opened 2025-06-26
memory/interprocess: Update boost to 1.87.0

umaucher:
Use the latest boost libs from the official bazel registry and its usage


score/os/utils/interprocess/BUILD
```diff
@@ -18,6 +18,8 @@ cc_library(
     srcs = ["interprocess_mutex.cpp"],
     hdrs = ["interprocess_mutex.h"],
     features = COMPILER_WARNING_FEATURES,
+    # ToDo: Remove the linkopts if clarified with score toolchain (needed after boost dep update)
+    linkopts = ["-lpthread"],
```
LittleHuba, line 22: This must be moved to the toolchain.
https://github.com/eclipse-score/toolchains_gcc/blob/702dd58815acfc8fbf80d71e8ae22470e1b6f682/toolchain/internal/cc_toolchain_config.bzl#L216

@AlexanderLanin @ltekieli could you please discuss adding this to the toolchain?

umaucher: @nradakovic ^^

nradakovic: It cannot just be moved to the toolchain. Whatever we put in the toolchain will have far reaching effects and as such each feature which is added (as default) to the TC, needs to be discuss & approved by architecture group (or whoever is responsible for making this kind of decisions) since other modules will be affected.
So questions which we need to consider are:
1. Does the flag/feature must be added in each module?
2. Do we need an option to disable/enable flag on config level?
3. Will there be any alternative to pthread?

LittleHuba: Agreed. This is something that we need answers to. So who would be the people we need for this discussion?

PR #14, opened 2025-08-15
Use system dependencies

tmewett:
This is a showcase of baselibs building with all system dependencies for all libraries and tools, for demonstration and discussion.

The S-CORE infrastructure team has been discussing how S-CORE can be integrated into a wider system. It is usually requirements for integrators that:

- they can control all dependencies and tools in a centralised manner, e.g. via a meta-build tool such as Bitbake or BuildStream
- no pre-built binaries are brought in by any component

Currently, S-CORE components do not meet either of these requirements, because:

- all dependencies are pinned to exact versions and pulled in by Bazel
- a specific pre-built compiler is downloaded
- pre-built libraries are downloaded from Ubuntu (though this one is just a temporary workaround)

This PR modifies baselibs to address these points. All dependencies now must be installed in the system in order to build. For me on Ubuntu, that means installing these packages:

`g++ libacl1-dev libcap-dev libseccomp-dev valgrind libboost-container1.83-dev nlohmann-json3-dev`

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -18,6 +18,6 @@ common --registry=https://bcr.bazel.build
 common --//score/mw/log/detail/flags:KUse_Stub_Implementation_Only=False
 common --//score/mw/log/flags:KRemote_Logging=False
 common --//score/json:base_library=nlohmann
-common --extra_toolchains=@gcc_toolchain//:host_gcc_12
 build --incompatible_strict_action_env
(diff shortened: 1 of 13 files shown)


opajonk:
> I guess this goes hand in hand with #15

Yes, probably. The devcontainer could supply these dependencies for "development host builds", of course.

tmewett:
By the way, my system GCC is version 13. I've been trying to build with 14 and 15 too, and they pick up a lot of new errors with baselibs, such missing #includes. So this PR will only build with GCC 13 and lower.

4og:
I'm closing this for now.
@nradakovic is working on a PoC on how users could customize parts of baselibs to their needs. See https://github.com/eclipse-score/baselibs/issues/305

PR #24, opened 2025-10-11
tracing(interface_types): make headers public and add trace_library_interface to generic_trace_api_header deps

Abdelaziz177:
**Visibility**: switch the following targets to //visibility:public so downstream repos (e.g. communication) can depend on them directly:

****generic_trace_api_header**
**trace_library_interface**
**meta_info**
**type_definitions****

**Dependency fix:** add :trace_library_interface to :generic_trace_api_header deps.
generic_trace_api.h includes i_trace_library.h, so this must be provided transitively; otherwise downstream builds see missing headers.

**Background / context**

While building the QNX IPC example and communication’s tracing code, we hit missing-header and visibility errors because
@score-baselibs//score/analysis/tracing/generic_trace_library/interface_types targets were restricted to __subpackages__
and generic_trace_api_header didn’t expose the interface include it uses.

communication consumes the interface types directly (not the full tracing lib), so these symbols need to be publicly visible.

**Impact**

Unblocks communication/score/mw/com/impl/tracing and the QNX IPC example build.

No behavioral changes; build/visibility only.

**Notes for reviewers**

Only BUILD file changes; no C++ source changes.

The change preserves existing warning features (COMPILER_WARNING_FEATURES).

**Testing**

Rebuilt locally after the change and verified that communication can depend on the interface targets without visibility or missing-header failures.

Built the QNX IPC example with the updated deps.

### DIFF
--- a/score/analysis/tracing/generic_trace_library/interface_types/BUILD
+++ b/score/analysis/tracing/generic_trace_library/interface_types/BUILD
@@ -25,80 +25,67 @@
 load("@score-baselibs//score/language/safecpp:toolchain_features.bzl", "COMPILER_WARNING_FEATURES")
 
 cc_library(
-name = "generic_trace_api_header",
-hdrs = ["generic_trace_api.h"],
-features = COMPILER_WARNING_FEATURES,
-tags = ["FFI"],
-
-
-
-visibility = ["//visibility:public"],
-deps = [
-    ":meta_info",
-    "@score-baselibs//score/analysis/tracing/generic_trace_library/interface_types/chunk_list",
-    "@score-baselibs//score/analysis/tracing/generic_trace_library/interface_types/error_code",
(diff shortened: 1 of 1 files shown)


score/analysis/tracing/generic_trace_library/interface_types/BUILD
```diff
@@ -80,7 +87,5 @@ cc_library(
     ],
     features = COMPILER_WARNING_FEATURES,
     tags = ["FFI"],
-    visibility = [
-        "@score_baselibs//score/analysis/tracing:__subpackages__",
-    ],
-)
+    visibility = ["//visibility:public"],
+)
```
hoppe-and-dreams: Please proper file ending according to [POSIX](https://stackoverflow.com/a/729795) is a new line.

4og:
This PR has no activity since a while, I'm closing.

PR #27, opened 2025-10-12
qnx: make mman_impl LP64-safe; use ::mmap/::mem_offset on QNX LP64 and include <sys/mman.h>, <sys/types.h>

Abdelaziz177:
**Problem**

When building on QNX LP64 (e.g., Neutrino 8 x86_64), the 64-suffixed APIs are often not declared because the non-suffixed ones already use 64-bit off_t. Calling ::mmap64 / ::mem_offset64 unconditionally produces errors like:

'::mmap64' has not been declared

'::mem_offset64' has not been declared

**What this PR does**

Adds the missing system headers <sys/mman.h> and <sys/types.h> at the top of mman_impl.cpp.

Uses feature checks to select the correct calls:

On __QNX__ && __LP64__, mmap64() delegates to ::mmap(...) (already 64-bit).

On __QNX__ && __LP64__, mem_offset64() delegates to ::mem_offset(...) (already 64-bit).

On all other platforms, behavior is unchanged (still calls the 64-suffixed versions).

**Why this is correct**

On LP64 systems, off_t is 64-bit, so the non-suffixed calls already use wide offsets.

QNX headers may not even declare the “64” symbols on LP64 because they are redundant.

### DIFF
--- a/score/os/qnx/mman_impl.cpp
+++ b/score/os/qnx/mman_impl.cpp
@@ -11,7 +11,8 @@
  * SPDX-License-Identifier: Apache-2.0
  ********************************************************************************/
 #include "score/os/qnx/mman_impl.h"
-
+#include <sys/mman.h>
+#include <sys/types.h>
 namespace score
 {
 namespace os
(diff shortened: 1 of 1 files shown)


4og:
Merged in https://github.com/eclipse-score/baselibs/commit/9dfcae31ffc7c9722d9cfd105be4e20a205ef1f4

PR #31, opened 2025-10-28
fix dependency to fsnotify lib

antonkri:


4og:
Fixed with https://github.com/eclipse-score/baselibs/commit/5d25c971a50fe5c97cd45b937762d5f1409feb6c

PR #33, opened 2025-10-29
Add CMake project

DominikSkanda:
Such that not only bazel is supported as build system I have added a basic CMake project structure, currently focusing on linux (not QNX) such that baselibs can be used by the lola communication module, where I also added a basic CMake project structure.


fbaeuerle:
Hi @DominikSkanda,
thank you for your work on baselibs. Unfortunately we cannot maintain two build systems in parallel.
Sorry for the disappointment and thank you for your understanding.

In case your PR contains other fixes that you'd still want to contribute, feel free to open a new PR :).

PR #51, opened 2025-12-02
testing: add first component integration test to baselib

PiotrKorkus:
initial version of CITs for baselibs


pahmann (changes requested):

pahmann (approved):
Only checked scope and that tests have a trace using described methods.
Improvement point is to improve the language of the test case descriptions/specification to indicate more clearly what it does, expected output, etc.
As this is the initial commit to "have something working". It is sufficiently described.
(Note: Functional check and integration was not performed)

4og (changes requested):
Hi @PiotrKorkus, could you help me understand the benefits of CITs over plain Google Tests? Since the baselibs consist of `cc_library` targets, they can easily be tested with Google Tests without the added complexity Python tooling of CIT.

anmittag:
@PiotrKorkus your PR is still relevant? when do you plan to fix the comments?

4og:
Feel free to reopen if still relevant

PR #52, opened 2025-12-03
Restrict access to MemoryResourceProxy in ManagedMemoryResource

jorgecasal:
This pull request removes the public visibility of `getMemoryResourceProxy()` in `ManagedMemoryResource` and changes the `PolymorphicOffsetPtrAllocator` constructor to accept `ManagedMemoryResource` instead of a proxy pointer. Only `PolymorphicOffsetPtrAllocator` is allowed to call `getMemoryResourceProxy()` via friend class access.

resolves [#94](https://github.com/eclipse-score/communication/issues/94)


fbaeuerle:
Hi @jorgecasal, this change looks great – Thanks!
I am keen to import your change into our internal repository via copybara and export it back to baselibs.
In order to do so, I would like to ask you to squash your commits together (or rebase them in a suitable way, the merge-commit is likely to cause trouble otherwise).

jorgecasal:
Hi @fbaeuerle, thanks for the feedback!
I've squashed the commits into a single one. Let me know if you need anything else.

PR #67, opened 2025-12-17
Example how to use the module api

ramceb:


score/concurrency/BUILD
```diff
@@ -467,3 +471,71 @@ cc_unit_test_suites_for_host_and_qnx(
...
+    name = "architectural_design",
+    srcs = ["design/architectural_design.md"],
+)
+
+sphinx_docs_library(
+    name = "safety_analysis",
+    srcs = ["design/safety_analysis.md"],
+)
+
+safety_element_out_of_context(
```
AlexanderLanin: Can you show how this would be used? e.g. by a different module or by integration?

score/concurrency/BUILD
```diff
@@ -467,3 +471,71 @@ cc_unit_test_suites_for_host_and_qnx(
...
+    name = "seooc_concurrency",
```
AlexanderLanin: content question: is concurrency a seeoc or baselibs?

PR #105, opened 2026-02-16
Codeowners comment adapted to be used from PMP - Org

anmittag:
contents wise no change, we just want to use the code owners file in future to document the score organisation, therefore added some comments


4og:
This looks not relevant anymore.

PR #108, opened 2026-02-22
remove unnecessary lock in SharedMemoryResource::do_deallocate

ShoroukRamzy:
### Summary

This PR removes an unnecessary `std::lock_guard` from the `SharedMemoryResource::do_deallocate` method.

### Problem

The `SharedMemoryResource` implements a monotonic (Arena) allocation strategy, where `deallocate` is a no-op. The current implementation acquires a heavyweight `InterprocessMutex` before doing nothing, which introduces unnecessary performance overhead on every deallocation call.

### Solution

The `std::lock_guard` is removed from the function. Since no shared state is read or modified, no lock is required.

### Benefits

-   **Performance:** Eliminates a costly and unnecessary system call, making deallocation a true zero-cost operation as expected from a monotonic allocator.
-   **Clarity:** The code now more accurately reflects the no-op nature of the function.


fbaeuerle:
This got integrated: https://github.com/eclipse-score/baselibs/commit/464808986a369be96bfee7c74b6310e695ce8bda

PR #110, opened 2026-02-25
const correctness for public interfaces in MemoryManagedResource

ShoroukRamzy:
**Description**

This pull request **solves this issue** (https://github.com/eclipse-score/communication/issues/84) and improves the API safety and robustness of the score::memory::shared library by enforcing const-correctness on key methods. **The primary goal is to prevent accidental modification of internal bookkeeping data within shared memory segments, which could lead to memory corruption.**

The following key changes have been implemented:

**const Return Types and Methods:**

**ManagedMemoryResource::getBaseAddress()** now returns const void*. This protects the underlying ControlBlock from being modified by external callers.
**ManagedMemoryResource::getMemoryResourceProxy()** now returns const MemoryResourceProxy* and is a const method. This ensures the integrity of the proxy's identifier.

**Derived Class and Mock Updates:**

All classes inheriting from ManagedMemoryResource or ISharedMemoryResource (including SharedMemoryResource, NewDeleteDelegateMemoryResource, and all test mocks/fakes) have been updated to correctly override the new const-correct signatures.

**Call Site Adjustments:**

All call sites that invoke getBaseAddress() or getMemoryResourceProxy() have been updated to correctly handle the const return types.
const_cast has been used in a few specific, controlled instances (primarily in test fakes interacting with the MemoryResourceRegistry) to maintain compatibility with older APIs that were not yet const-aware. This was a pragmatic choice to limit the scope of the refactoring while still achieving the primary safety goal.

**How to Test**

**Build the score/memory library and its dependencies:**
bash
`bazel build --config=bl-x86_64-linux //score/memory/...`
**Run all associated tests to ensure no regressions have been introduced:**
bash
`bazel test --config=bl-x86_64-linux //score/memory/...`
All tests should pass, confirming that the const-correctness changes have been applied consistently and correctly throughout the library and its test suite.

### DIFF
--- a/score/analysis/tracing/common/interface_types/shared_memory_location_helpers.h
+++ b/score/analysis/tracing/common/interface_types/shared_memory_location_helpers.h
@@ -62,7 +62,7 @@ auto GetPointerFromLocation(SharedMemoryLocation memory_location, ResourcePointe
     // coverity[autosar_cpp14_m5_0_17_violation]
     // coverity[autosar_cpp14_a5_2_4_violation]
     // coverity[autosar_cpp14_m5_2_8_violation]
-    return reinterpret_cast<T*>(static_cast<std::uint8_t*>(resource_ptr->getBaseAddress()) + memory_location.offset_);// NOLINT(cppcoreguidelines-pro-type-reinterpret-cast, cppcoreguidelines-pro-bounds-pointer-arithmetic): Tolerated see above
+    return reinterpret_cast<T*>(const_cast<std::uint8_t*>(static_cast<const std::uint8_t*>(resource_ptr->getBaseAddress())) + memory_location.offset_); // NOLINT(cppcoreguidelines-pro-type-reinterpret-cast, cppcoreguidelines-pro-bounds-pointer-arithmetic): Tolerated see above
     // clang-format on
 }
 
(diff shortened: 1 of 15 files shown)


ShoroukRamzy:
> Hi @ShoroukRamzy, thank you for your contribution. Can you please rebase your commits? There have been changes in `score/memory/shared` in [1198d85](https://github.com/eclipse-score/baselibs/commit/1198d855e642bbc425ef6c32ce132ebd85d5a92a)

Hi @4og, Thank you for your feedback. Done!

fbaeuerle:
Hi @ShoroukRamzy,
Thank You for this contribution! I'm integrating this into score-baseslibs. It may need some adaptations in our internal repo (we import it there via copybara and then export it back to baselibs) and it also needs to go through an internal review. It may take a while and I'll update you as soon as its done.

fbaeuerle:
Sorry for letting this hang around for so long without any update.

We're currently in the situation where communication is integrating via github and baselibs (mostly) is mostly integrating via an internal repository (which exports those changes via copybara), which makes it hard to bring your change. Apart from that, communication is also in the process of migrating the code in memory/shared from baselibs to communication. It might make sense to give it another try when the code lands there.

Therefore I'm closing this PR here.

ShoroukRamzy:
> Sorry for letting this hang around for so long without any update.
> 
> We're currently in the situation where communication is integrating via github and baselibs (mostly) is mostly integrating via an internal repository (which exports those changes via copybara), which makes it hard to bring your change. Apart from that, communication is also in the process of migrating the code in memory/shared from baselibs to communication. It might make sense to give it another try when the code lands there.
> 
> Therefore I'm closing this PR here.

Hi @fbaeuerle and @crimson11, since shared memory is currently integrated in the mw::com repo. Can we integrate this work there somehow or what do you think? Thanks!

PR #113, opened 2026-03-14
test(os/time): add coverage for pid 0, TIMER_ABSTIME, and ovalue

nhuvaoanh123:
Three gaps identified in time_test.cpp:

1. ClockGetCpuClockIdWithCurrentProcess: Existing test uses pid 1 (init), which is semantically incorrect and may fail on privilege-restricted systems. POSIX specifies pid 0 means the calling process — more portable and spec-correct.

2. TimerSettimeWithAbsoluteTimeFlag: All existing timer_settime calls use flags=0 (relative). TIMER_ABSTIME is a distinct code path (absolute wall-clock timestamp interpretation) with no prior test coverage.

3. TimerSettimeOldValueReflectsPreviousArming: TimerSettimeSuccess passes an ovalue but never asserts its contents. This test re-arms a live timer and verifies ovalue correctly reports the remaining time of the previous arming (~10s).

All tests annotated with SCR-46010294, ASIL B, Interface test.

 Note for future improvement:
Test 1: pid=0 is Linux-only — QNX CI may fail; assertion is too shallow
Test 2: 200ms absolute deadline is flaky under sanitizers; missing SIGEV_NONE could kill the test process
Test 3: The 8–10s range window is the most likely reviewer flag — asan/ubsan slowdown can blow this
All three: No negative/error path coverage (ASIL B typically requires this); RecordProperty may not be consumed by baselibs CI

### DIFF
--- a/score/os/test/time_test.cpp
+++ b/score/os/test/time_test.cpp
@@ -531,6 +531,120 @@ TEST(TimeImplTest, TimerWithSigEvSignalNotificationIsSuccessful)
     Time::instance().timer_delete(timerid);
 }
 
+TEST(TimeImplTest, ClockGetCpuClockIdWithCurrentProcess)
+{
+    RecordProperty("Verifies", "SCR-46010294");
+    RecordProperty("ASIL", "B");
+    RecordProperty("Description", "Verifies Clock Get Cpu Clock Id Success With Current Process Pid Zero");
+    RecordProperty("TestType", "Interface test");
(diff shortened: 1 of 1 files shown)


Rutik7:
Hi @nhuvaoanh123 
This was integrated 
https://github.com/eclipse-score/baselibs/commit/ea4d9e03b4fa7a969c4a9c5933f8422e3a3e22ea

PR #120, opened 2026-03-24
replaced score::cpp with std

marwan275:
This PR migrates tracing and related components from  score::cpp types to standard C++17 std equivalents.


4og:
Closing as this change is irrelevant after refactoring in analysis/tracing

PR #134, opened 2026-03-31
Enable QNX UT for x86-64

ltekieli:


ltekieli:
See: https://github.com/eclipse-score/baselibs/pull/140

PR #136, opened 2026-04-02
creating pull request for issue:2628

balajiyalam2:
updated workflows to throw error if MODULE.bazel.lock is missing (or) not in sync with MODULE.bazel file


4og:
There is another approach to validate the lockfile: #252

PR #139, opened 2026-04-08
Add vajson to score baselibs

### LINKED ISSUE #201: Integrate vajson with parsing functionality into Baselibs
### What

vajson librariry with parsing functionality is to be added to baselibs

### Acceptance Criteria (DoD)

- [ ] The code is placed in `score/json/internal/parser/vajson`

JuriSchroeder:
This pull request adds the vector VaJson Parser to the score baselibs.
Closes #201


JuriSchroeder:
@FScholPer 
I removed the amsr namespace and replaced it with score::json::vajson. I also updated the header and source file structure to reflect that and be more inline with the rest of the repository. 
Some tests are still failing on aarch64, but passing on x86_64. I'll investigate.

JuriSchroeder:
@mihajlo-k @fbaeuerle This PR is ready for review. The failing copyright check is pointing to a file that is not changed as part of this PR. Let me know if something needs changing.

.bazelrc
```diff
@@ -20,7 +20,7 @@ build --java_runtime_version=remotejdk_17
 build --tool_java_runtime_version=remotejdk_17
 build --credential_helper="*.qnx.com=%workspace%/.github/tools/qnx_credential_helper.py"
 
-build:_bl_stub --//score/json:base_library=nlohmann
+build:_bl_stub --//score/json:base_library=vajson
```
fbaeuerle: Does this change need project alignment?

score/json/internal/parser/vajson/vajson_impl/reader/internal/parsers/structure_parser_base.cpp
```diff
@@ -405,8 +405,10 @@ auto StructureParserBase::ParseLength() noexcept -> Result<std::uint32_t>
       .ReadExactly(kPrefixSize,
                    [&result](StringView view) noexcept {
                      static_cast<void>(std::memcpy(&result, view.data(), sizeof(result)));
-
-                     result = be32toh(result);
```
fbaeuerle: out of curiosity: why can't we use be32toh?

fbaeuerle:
> @mihajlo-k @fbaeuerle This PR is ready for review. The failing copyright check is pointing to a file that is not changed as part of this PR. Let me know if something needs changing.

The problem is anyway already fixed on master. I cannot rerun it (neither can I approve the qnx workflows).
I'd like to verify the build running on qnx before merging this

score/json/internal/parser/vajson/vajson_impl/reader/internal/parsers/structure_parser_base.cpp
JuriSchroeder: IIRC it caused build failures on QNX because the macro is not defined there.

.bazelrc
JuriSchroeder: I guess so. I can also just revert it for now.

JuriSchroeder: @fbaeuerle I switched back to using nlohmann as default.

mihajlo-k (commented):
> **Reference File:** [`score/json/json_parser.cpp` Lines 27-35](https://github.com/JuriSchroeder/baselibs/blob/8b7d5a5be63f70403a253a0ce94347378facde0e/score/json/json_parser.cpp#L27-L35)
```cpp
auto JsonParser::FromFile(const std::string_view file_path) const noexcept -> score::Result<Any>
{
    return NlohmannParser::FromFile(file_path);
}

auto JsonParser::FromBuffer(const std::string_view buffer) const noexcept -> score::Result<Any>
{
    return NlohmannParser::FromBuffer(buffer);
}
```
here it's still calling `NlohmannParser` unconditionally. maybe we should add some `#ifdef VAJSON` logic in this PR or something similar?

JuriSchroeder:
@mihajlo-k 
I moved the implementation of the FromFile and FromBuffer functions into the corresponding libraries. This way the correct implementation is chosen by setting @score_baselibs//score/json:base_library=nlohmann or @score_baselibs//score/json:base_library=vajson

score/json/internal/parser/vajson/vajson_impl/util/json_error_domain.h
```diff
@@ -0,0 +1,289 @@
...
+using CStr = const char*;
+
+/// \brief            Unqualified access to Result
+/// \tparam          T                  Type of value.
+template <typename T>
+using Result = score::Result<T>;
+
+using Blank = score::cpp::blank;
+
+using ResultBlank = score::Result<score::cpp::blank>;
```
mihajlo-k: you can safely use `score::Result` directly. The semantics of the `ResultBlank` changed in the meantime from `Result<score::cpp::blank>` to `Result<void>`. However, it is desired that we use `Result<void>` directly instead of `ResultBlank` to avoid misinterpretation, especially since it had different meaning before.

mihajlo-k:
> @mihajlo-k I moved the implementation of the FromFile and FromBuffer functions into the corresponding libraries. This way the correct implementation is chosen by setting @score_baselibs//score/json:base_library=nlohmann or @score_baselibs//score/json:base_library=vajson

@JuriSchroeder I had this in mind: https://github.com/eclipse-score/baselibs/pull/214. this makes specific implementation more self-contained since they don't include higher-level `json_parser.h`. On the other hand, in your approach there is no preprocessor conditional compilation...

so, I am fine either way.

JuriSchroeder:
@mihajlo-k 
I'm not sure to what extent the AUTOSAR C++14 guidelines are applicable to S-CORE, but I've seen some justifications in the code base so I assume that it is somewhat relevant.

Rule A16-0-1 states:
`The pre-processor shall only be used for unconditional and conditional file
inclusion and include guards, and using the following directives: (1) #ifndef,
(2) #ifdef, (3) #if, (4) #if defined, (5) #elif, (6) #else, (7) #define, (8) #endif, (9)
#include.`
If it's OK I'd go with my solution, even if it has the drawback of including the higher-level json_parser.h

mihajlo-k:
> @mihajlo-k I'm not sure to what extent the AUTOSAR C++14 guidelines are applicable to S-CORE, but I've seen some justifications in the code base so I assume that it is somewhat relevant.
> 
> Rule A16-0-1 states: `The pre-processor shall only be used for unconditional and conditional file inclusion and include guards, and using the following directives: (1) #ifndef, (2) #ifdef, (3) #if, (4) #if defined, (5) #elif, (6) #else, (7) #define, (8) #endif, (9) #include.` If it's OK I'd go with my solution, even if it has the drawback of including the higher-level json_parser.h

@JuriSchroeder yes, it's fine. please proceed with your solution. I closed: https://github.com/eclipse-score/baselibs/pull/214

JuriSchroeder:
@mihajlo-k 
I removed every use of Blank and ResultBlank and replaced it with Result<void>

mihajlo-k:
> @mihajlo-k I removed every use of Blank and ResultBlank and replaced it with Result

@JuriSchroeder looks good! thanks!

PR #140, opened 2026-04-08
Add QNX tests to CI workflow

4og:


third_party/itf/py_unittest_qnx_test.bzl
```diff
@@ -12,4 +12,8 @@
 # *******************************************************************************
 
 def py_unittest_qnx_test(**kwargs):
+    # This is intentionally left blank.
```
kitsnet, line 15: So, what do we do with the existing test targets using this rule in eclipse-score/communication, for example? How do we convert them so that they start running in CI?

4og: It's either you use `cc_test_qnx` directly or run your tests via the run_under approach.
See example here: https://github.com/eclipse-score/qnx_unit_tests?tab=readme-ov-file#usage

Another option is to implement `py_unittest_qnx_test` in eclipse-score/communication as in https://github.com/eclipse-score/baselibs/pull/134/changes#diff-f8b9491af30d090714acefd4e798a7582f992ae1546f566c6ccc2674161efefdR15

I'd avoid using the `py_unittest_qnx_test` from baselibs, because it would require the `score_qnx_unit_tests` to be turned to a non-dev-dependency.

4og:
Merged in https://github.com/eclipse-score/baselibs/commit/ef2a4fd5cbb3fcc29e65fb70718c91e1fccade8c

PR #144, opened 2026-04-10
Change name of nlohmann_json to score_nlohmann_json

ThomasClausnitzer:
This PR prepares Issue https://github.com/eclipse-score/baselibs/issues/242. 

It fulfills the following task:
- Adding the nlohmann_json to the [S-CORE bazel registry](https://github.com/eclipse-score/bazel_registry). The module name should be prefixed with score_. I.e. score_nlohmann_json.

**Note:** It currently uses git_override to point to the PR branch https://github.com/eclipse-score/nlohmann_json/pull/32. This is because there is no official release of nlohmann_json for bazel yet. When that release it created AND PR32 is approved, this line can and should be removed.


anmittag:
@ThomasClausnitzer several checks for your PR fail. are you going to correct it or remove the PR?

PR #170, opened 2026-04-24
Add GitHub workflow for testing baselibs integration.

ssingh2099:
Add a minimal integration example and CI check to ensure baselibs remains consumable as a Bazel module in an external workspace.


4og (approved):

examples/integration/MODULE.bazel
```diff
@@ -0,0 +1,28 @@
...
+# information regarding copyright ownership.
+#
+# This program and the accompanying materials are made available under the
+# terms of the Apache License Version 2.0 which is available at
+# https://www.apache.org/licenses/LICENSE-2.0
+#
+# SPDX-License-Identifier: Apache-2.0
+# *******************************************************************************
+
+bazel_dep(name = "score_baselibs", version = "0.0.0")
```
sankurm, line 14: Incorrect Version: Is it intentional that the example will always only use version 0.0.0? Shouldn't the example of any branch (which is `main` here) build its own version?

Unnecessary dependency: In fact, wouldn't the `local_path_override` below use the current branch making this dependency redundant?

examples/integration/main.cpp
```diff
@@ -0,0 +1,19 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+// *******************************************************************************
+
+#include "score/result/result.h"
+
+int main(int, char**)
+{
+    return 0;
+}
```
sankurm, line 19: No functionality: `result.h` is included but the program does nothing - not even an instantiate a `result<int>`. Is this intentional? Is the expected functionality defined in a work item (Issue/Bug)? Sorry, I am unable to find any linked item for this PR to determine what this program is expected to do.

examples/integration/README.md
```diff
@@ -0,0 +1,48 @@
...
+This example serves as an integration test in CI to verify that Base Libraries:
+- Can be consumed as a module by external projects
+- Has correct transitive dependency declarations
+- Exposes only intended public targets
+
+## Building the Example
+
+```bash
+cd examples/integration
+bazel build //...
```
sankurm, line 16: Correctness: Is this the right command to build the integration example or does it build all of baselibs? In my opinion, this command should build just the example, as the section title suggests.

examples/integration/README.md
```diff
@@ -0,0 +1,48 @@
...
+```
+
+## Modifying the Example
+
+To add additional base libraries features to test:
+
+1. Add dependencies to the `cc_binary` target in `BUILD`
+2. Include headers in `main.cpp`
+3. Run `bazel build //...` to verify
```
sankurm, line 25: Repetition: The same command as in the `Building the Example` section is repeated here. This redundant information may result in an inadvertent future deviation, and the 2 places may specify different commands confusing the user. Kindly consider referring to the other section.

.gitignore
```diff
@@ -1,4 +1,5 @@
 bazel-*
+examples/integration/bazel-*
```
sankurm, line 2: Question: Wouldn't the previous line already ignore this already?

.github/workflows/build_linux.yml
```diff
@@ -68,6 +68,11 @@ jobs:
       - name: Bazel Test
         run: |
           bazel test --config ${{ matrix.bazel-config }} ${{ matrix.bazel-test-extra-flags }} -- ${{ matrix.bazel-test-target }}
+      - name: Bazel test integration as module
+        if: matrix.bazel-config == 'bl-x86_64-linux'
+        run: |
+          # Test that baselibs works when consumed as an external module dependency
+          cd examples/integration && bazel build //... && bazel mod deps --lockfile_mode=update
```
sankurm, line 75: Unnecessary command: Is the `bazel mod deps` command really needed here?

examples/integration/.bazelversion
```diff
@@ -0,0 +1 @@
+8.3.1
```
sankurm, line 1: Overspecification: Do we really need this specified for the example? Shouldn't the `MODULE.bazel` specify this? Remove this file?

sankurm (commented):
I am sharing my thoughts late. I am happy to discuss any of the points.

.github/workflows/build_linux.yml
ssingh2099: Yes, this is intentional. We need this step to validate the baselibs module dependency graph and ensure all dependencies resolve correctly.

examples/integration/.bazelversion
ssingh2099: MODULE.bazel manages dependencies, while .bazelversion pins the Bazel tool version for consistent local and CI builds. It is not strictly required for functionality, so I am also fine removing it if we prefer to keep the example minimal.

examples/integration/main.cpp
ssingh2099: the example is a minimal CI smoke test to verify external module consumption and dependency build, not a full functional showcase.

examples/integration/MODULE.bazel
ssingh2099: 0.0.0 is a placeholder because the dependency is overridden to the local checkout, so the example builds the current branch. 
bazel_dep is still required to declare the dependency, local_path_override only changes where it is resolved from.

examples/integration/README.md
ssingh2099: Yes, bazel build //... here builds all targets in integration, not the full baselibs repository.

examples/integration/README.md
ssingh2099: Agreed. We will keep the build command in a single place and link to the "Building the Example" section here to avoid duplication

.gitignore
ssingh2099: You’re right that bazel-* would usually cover this. Added the explicit ignore because Bazel was following symlinks in examples/integration/bazel-* and getting stuck in an infinite loop. This path-specific rule prevents that.

examples/integration/MODULE.bazel
sankurm: Okay. Thanks!

examples/integration/main.cpp
sankurm: Is there a definition of what the example should finally look like? I am interested in seeing that please.

.gitignore
sankurm: Thanks for the clarification. (1) Is the problem with the infinite loop a Bazel problem or it is about the contents of our directory? (2) Is this the only way to address the issue?

examples/integration/.bazelversion
sankurm: If we keep this, the least version that can support our needs should be here. Maybe 8.3.1 is that? Or can we go lower?

PR #174, opened 2026-05-05
flatbuffers: Add opt-in buffer identification mechanism

### LINKED ISSUE #233: Docs generation warning: invalid requirements links referencing `comp__baselibs_flatbuffers` in flatbuffers docs
There are documentation generation warnings that will become fatal in the future. The following references in the flatbuffers docs link to invalid needs:

- baselibs/components/flatbuffers/docs/requirements/index.rst:112: comp_req__flatbuffers__serialization — links to invalid need(s): {'comp__baselibs_flatbuffers'}
- baselibs/components/flatbuffers/docs/requirements/index.rst:127: comp_req__flatbuffers__access — links to invalid need(s): {'comp__baselibs_flatbuffers'}
- baselibs/components/flatbuffers/docs/requirements/index.rst:141: comp_req__flatbuffers__verification — links to invalid need(s): {'comp__baselibs_flatbuffers'}
- baselibs/components/flatbuffers/docs/requirements/index.rst:159: comp_req__flatbuffers__buffer_identification — links to invalid need(s): {'comp__baselibs_flatbuffers'}
- baselibs/components/flatbuffers/docs/requirements/index.rst:171: comp_req__flatbuffers__version_check — links to invalid need(s): {'comp__baselibs_flatbuffers'}
- baselibs/components/flatbuffers/docs/requirements/index.rst:186: comp_req__flatbuffers__asil — links to invalid need(s): {'comp__baselibs_flatbuffers'}

See this job run for details: [GitHub Actions job run](https://github.com/eclipse-score/baselibs/actions/runs/27016884202/job/79738940934)

### LINKED ISSUE #234: Docs generation warning: invalid requirements links referencing `comp__baselibs_vajson` in vajson docs
There are documentation generation warnings that will become fatal in the future. The following references in the vajson docs link to invalid needs:

- baselibs/components/json/docs/vajson/requirements/index.rst:32: comp_req__vajson__validation — links to invalid need(s): {'comp__baselibs_vajson'}
- baselibs/components/json/docs/vajson/requirements/index.rst:44: comp_req__vajson__trailing_commas — links to invalid need(s): {'comp__baselibs_vajson'}
- baselibs/components/json/docs/vajson/requirements/index.rst:56: comp_req__vajson__hex_integers — links to invalid need(s): {'comp__baselibs_vajson'}
- baselibs/components/json/docs/vajson/requirements/index.rst:68: comp_req__vajson__unicode — links to invalid need(s): {'comp__baselibs_vajson'}
- baselibs/components/json/docs/vajson/requirements/index.rst:82: comp_req__vajson__asil — links to invalid need(s): {'comp__baselibs_vajson'}

See this job run for details: [GitHub Actions job run](https://github.com/eclipse-score/baselibs/actions/runs/27016884202/job/79738940934)

OliverHeilwagen:
resolves #128 
fixes #233
fixes #234


score/flatbuffers/bazel/tools.bzl
```diff
@@ -222,10 +245,194 @@ serialize_multiple_buffers = rule(
...
+            default = "//score/flatbuffers/bazel:inject_buffer_version",
+            executable = True,
+            cfg = "exec",
+            doc = "The inject_buffer_version helper script for JSON version injection.",
+        ),
+    },
+    doc = "Injects major/minor version fields into a JSON file for use with serialize_versioned_buffer.",
+)
+
+def serialize_versioned_buffer(
```
OliverHeilwagen, line 310: @4og @paulquiring 
I introduced the following rules separately so we can have the major/minor version as mandatory rule parameter:
- serialize_versioned_buffer
- serialize_multiple_versioned_buffer

I am not so happy with the longer name but kept it for now to avoid breaking changes.
Since we are still early with the flatbuffers rollout we could do a breaking change of the rule name and have the following ones that should remain stable:
- serialize_buffer
- serialize_buffers (breaking former: serialize_multiple_buffers)
- serialize_versioned_buffer
- serialize_versioned_buffers

The buffer version have dedicated `data` and `output` parameters.
The buffer**s** version have `data_dict` parameter (dictionary key: json data, value: output name).

Please let me know if you have any preferences.

score/flatbuffers/test/version_reader_test.cpp
```diff
@@ -0,0 +1,354 @@
...
+    score::flatbuffers::VersionReader reader_;
+
+    /// All tests access functionality through the abstract interface,
+    /// mirroring how production code should depend on IVersionReader.
+    score::flatbuffers::IVersionReader& version_reader_{reader_};
+};
+
+TEST_F(VersionReaderTest, GetVersionRejectsNullPointer)
+{
+    RecordProperty("PartiallyVerifies", "comp_req__flatbuffers__buffer_identification");
```
OliverHeilwagen, line 113: Check if process is considering multiple PartiallyVerifies as FullyVerifies, or if the test cases should already claim FullyVerifies.

score/flatbuffers/details/buffer_version_info.cpp
```diff
@@ -0,0 +1,38 @@
+/********************************************************************************
```
OliverHeilwagen, line 1: Correct name to buffer_version_info.cpp

score/flatbuffers/buffer_version_info.hpp
```diff
@@ -0,0 +1,87 @@
...
+/// `IVersionReader::VerifyVersion` must have
+/// `version_info:score.flatbuffers.BufferVersion` as field 0 of their root
+/// table, with version values explicitly stored in the binary.
+/// This is guaranteed when the buffer was produced by e.g. the
+/// `serialize_versioned_buffer` Bazel rule.
+class BufferVersionInfo
+{
+  public:
+    /// Number of characters in a FlatBuffers file identifier.
+    static constexpr std::size_t kIdentifierLength = 4U;
```
paulquiring, line 42: static_cast<std::size_t>(::flatbuffers::kFileIdentifierLength) instead of 4U?

score/flatbuffers/details/version_reader.cpp
```diff
@@ -0,0 +1,191 @@
...
+    // Read the file_identifier from bytes 4-7
+    char id[BufferVersionInfo::kIdentifierLength + 1U]{};
+    std::memcpy(id, data + sizeof(::flatbuffers::uoffset_t), ::flatbuffers::kFileIdentifierLength);
+
+    return BufferVersionInfo{id, version->major_version(), version->minor_version()};
+}
+
+inline score::Result<BufferVersionInfo> GetVersionImpl(const std::vector<uint8_t>& buf) noexcept
+{
+    if (buf.size() > 0)
```
paulquiring:
```suggestion
    if (!buf.empty())
```

score/flatbuffers/details/version_reader.cpp
```diff
@@ -0,0 +1,191 @@
...
+    // type_marker presence and correct value ensure no accidential match with a similar
+    // table based field.
+    const auto* marker = version->type_marker();
+    if ((marker == nullptr) || (std::string_view{marker->c_str()} != "BV"))
+    {
+        return MakeUnexpected(ErrorCode::kVerificationFailed);
+    }
+
+    // Read the file_identifier from bytes 4-7
+    char id[BufferVersionInfo::kIdentifierLength + 1U]{};
```
paulquiring: I'm happy to see the value-Initialization of id :)

score/flatbuffers/details/version_reader.cpp
```diff
@@ -0,0 +1,191 @@
...
+    std::memcpy(id, data + sizeof(::flatbuffers::uoffset_t), ::flatbuffers::kFileIdentifierLength);
```
paulquiring: Can you leave a comment why sizeof(::flatbuffers::uoffset_t) is needed?

score/flatbuffers/version_reader.hpp
```diff
@@ -0,0 +1,107 @@
...
+/// `version_info:score.flatbuffers.BufferVersion` as field 0).
+///
+class VersionReader : public IVersionReader
+{
+  public:
+    VersionReader() = default;
+    ~VersionReader() override = default;
+
+    /// @copydoc IVersionReader::GetVersion(const uint8_t*, std::size_t)
+    score::Result<BufferVersionInfo> GetVersion(const uint8_t* data, std::size_t size) noexcept override;
```
paulquiring: Here an in other API const uint8_t* data, std::size_t size can be replaced with std::span<const uint8_t> data

score/flatbuffers/test/serialize_buffer_rules/BUILD
```diff
@@ -12,7 +12,7 @@
 # *******************************************************************************
```
paulquiring, line 12: Did I just miss it, or is there no .fbs-test that actually tries out "or use the `(id: 0)` attribute explicitly"?

paulquiring (approved):
The implementation looks good. I had some small findings and left some comments.

score/flatbuffers/details/version_reader.cpp
OliverHeilwagen: FlatBuffers is providing `GetBufferIdentifier` i will update the code so we don't reimplement part of its logic.

score/flatbuffers/version_reader.hpp
OliverHeilwagen: Changed to score::cpp::span.

score/flatbuffers/test/serialize_buffer_rules/BUILD
OliverHeilwagen: Added test_versioned_explicit_ids.fbs and added to parameterized test in version_reader_test.cpp.

score/flatbuffers/buffer_version_info.hpp
OliverHeilwagen: In the header i wanted to avoid dependency to flatbuffers.hpp just for this one value equality is asserted in implementation.

examples/integration/flatbuffers/config_usecase/BUILD
```diff
@@ -13,7 +13,7 @@
 
 load("@rules_cc//cc:defs.bzl", "cc_test")
```
OliverHeilwagen, line 14: @4og Shall we move this to e.g. `examples/integration/flatbuffers/config_usecase`?

score/flatbuffers/test/version_reader_test.cpp
OliverHeilwagen: testcovered attribute on requirement is needed.
If this states YES than the PartiallyVerifies links are expected to fully verify the requirement, which is the intention.

score/flatbuffers/test/version_reader_test.cpp
```diff
@@ -0,0 +1,543 @@
...
+
+    const auto& info = result.value();
+
+    EXPECT_EQ(info.identifier(), param.version_info.identifier());
+    EXPECT_EQ(info.major_version, param.version_info.major_version);
+    EXPECT_EQ(info.minor_version, param.version_info.minor_version);
+}
+
+// clang-format off
+/// Equivalence classes for `VerifyVersion`
```
OliverHeilwagen, line 484: @aschemmel-tech
Note:
I see such tables more useful for a developer and accessor, instead of the scattered test specification approach which is [currently defined in the process](https://eclipse-score.github.io/process_description/main/process_areas/verification/guidance/verification_specification.html#test-description).

I've briefly aligned this topic with @pahmann.
The PR diff view is unfortunately the drawback, especially when using the split view, otherwise it should simplify the review for e.g. `equivalence-classes` and `boundary-values`.

examples/integration/flatbuffers/config_usecase/BUILD
4og: That's a good question. I have a gut feeling that it's better to keep examples close to the code they show case. So the current location is ok. Otherwise, it would be hard for a user to find corresponding example.
@fbaeuerle, do you have an opinion?

4og:
@OliverHeilwagen, can you please update the branch in you fork to the latest baselibs main?

examples/integration/flatbuffers/config_usecase/BUILD
OliverHeilwagen: The benefits from my view point would be:
- detect bazel visbility issues early, e.g. relative bazel targets in starlark rules/macros.
- copy/paste-able examples, referencing with @score_baselibs is not longer optional.

score/flatbuffers/bazel/tools.bzl
paulquiring: it current state is good.

examples/integration/flatbuffers/config_usecase/BUILD
fbaeuerle: @4og I do not have a strong opinion, looking at https://github.com/eclipse-score/module_template/ i only see the top-level examples directory, and looking at the examples directory in baselibs we would probably need a directory structure resembling the score/ directory to keep it organized.

Do we have guidelines on repo structures in score the same way we have them in BMW (like the enhancement proposals or engineering standard docs)?

paulquiring (approved):

examples/integration/flatbuffers/config_usecase/BUILD
OliverHeilwagen: In S-CORE the following guideline is available besides module_template:
https://eclipse-score.github.io/score/main/contribute/general/folder.html#module-folder-structure

However this only considers documentation (`docs/manuals/examples/`, `score/<component>/docs/manuals/`), there is no statement about executable examples, which would be strange to have under `docs/`.

4og: @OliverHeilwagen, if you prefer to convert this test to a black-box test (i.e. it would depend on the `@score_baselibs` module), then I agree, the `examples/integration` is a better place.

OliverHeilwagen: @4og Yes intention is black-box testing and simulate how another module is consuming it. I will move it to `examples/integration/flatbuffers/config_usecase`, so we can later differentiate it from other examples.

examples/integration/.bazelrc
```diff
@@ -17,3 +17,5 @@ common --registry=https://bcr.bazel.build
 # Required baselibs configuration settings
 common --@score_baselibs//score/json:base_library=nlohmann
 common --@score_baselibs//score/memory/shared/flags:use_typedshmd=False
+
+build --cxxopt=-Wno-error=deprecated-declarations
```
4og, line 21: We should start using score_bazel_cpp_toolchains in the examples. It provides this compile option out of the box. Follow-up #363

4og (commented):
I did only a shallow review. If @paulquiring is happy with the change, I'll merge it.

score/flatbuffers/bazel/codegen.bzl
```diff
@@ -110,23 +119,59 @@ generate_cpp = rule(
...
+        ),
     },
     doc = """Generates a C++ header file from a FlatBuffer schema (.fbs) file.
-    
+
     This rule uses the flatc compiler from the @flatbuffers repository with
     absolute paths, making it usable outside of this repository.
-    
+
+    @score_baselibs//score/flatbuffers/common:buffer_version.fbs is always included
```
paulquiring, line 144: This is important information. Having that inside user documentation would be helpful. I think we should crate an issue to put this into the user documentation.

paulquiring (approved):
I left a comment about starting user documentation for the opt-in buffer identification. We can bring this in with another PR.

score/flatbuffers/bazel/codegen.bzl
paulquiring: Issue created https://github.com/eclipse-score/baselibs/issues/386

examples/integration/.bazelrc
paulquiring: @4og I'm happy, lets merge it.

examples/integration/flatbuffers/config_usecase/demo_versioned.fbs
```diff
@@ -0,0 +1,85 @@
...
+table BinaryData {
+    value:[uint8] (id: 0);
+}
+
+// Root configuration object
+root_type MyComponentConfig;
+
+// Component specific file identifier – distinct from the unversioned "DEMO"
+// so that VerifyVersion can distinguish the two buffer formats.
+file_identifier "DEMO";
```
fbaeuerle, line 84: the comment says distinct form the unversioned "DEMO", which i assume corresponds to demo.fbs. demo.fbs, however uses the exact same file_identifier. Is that intentional?

score/flatbuffers/details/version_reader_impl.hpp
```diff
@@ -0,0 +1,107 @@
...
+    // Reject buffers where version_info was placed in the wrong vtable slot (user fbs).
+    // The FlatBuffers verifier is type-blind and would otherwise accept another
+    // field's bytes as a valid (all-default) BufferVersion table.
+    // type_marker presence and correct value ensure no accidental match with a similar
+    // table-based field.
+    const auto* marker = version->type_marker();
+    // NOTE: The "BV" literal must be kept in sync with the type_marker value in
+    // inject_buffer_version.py. Changing it here without updating that script
+    // (or vice versa) will cause buffer version validation to fail at runtime.
+    if ((marker == nullptr) || (std::string_view{marker->c_str()} != "BV"))
```
fbaeuerle: rather minor, but it seems like flatbuffer strings come with a string_view implementation (which would probably save one strlen):
https://github.com/google/flatbuffers/blob/master/include/flatbuffers/string.h#L31

score/flatbuffers/common/buffer_version.fbs
```diff
@@ -0,0 +1,62 @@
...
+
+/// Open for extension: additional fields may be appended in future revisions.
+/// FlatBuffers forward/backward compatibility guarantees that existing readers
+/// will safely ignore unknown fields added here.
+table BufferVersion {
+    /// Type marker used to detect wrong-field placement at runtime.
+    ///
+    /// The FlatBuffers verifier is structurally type-blind: it checks only that
+    /// offsets are in-bounds and sizes are valid, not that the data at a given
+    /// vtable slot actually belongs to the expected table type. A `required`
```
fbaeuerle: for clarification: the doc mentions the field is required, but the schema doesn't add the [required](https://flatbuffers.dev/schema/#3-required) attribute to it. is that intentional?

score/flatbuffers/details/version_reader.cpp
```diff
@@ -0,0 +1,140 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ********************************************************************************/
+
+#include "score/flatbuffers/version_reader.hpp"
+#include "score/flatbuffers/details/version_reader_impl.hpp"
+
+#include <vector>
+
+static_assert(score::flatbuffers::BufferVersionInfo::kIdentifierLength == ::flatbuffers::kFileIdentifierLength,
```
fbaeuerle, line 19: wouldn't it be clearer if the score::flatbuffers::BufferVersionInfo::kIdentifierLength was using the value from ::flatbuffers::kFileIdentifierLength right away?

fbaeuerle: Ah, I just found out in the header that this was intentional

score/flatbuffers/common/buffer_version.fbs
```diff
@@ -0,0 +1,62 @@
...
+/// 1. Include this file at the top of your schema:
+///
+///      include "buffer_version.fbs";
+///
+/// 2. Declare `version_info` as the FIRST field of your root table:
+///
+///      table MyConfig {
+///          version_info:score.flatbuffers.BufferVersion;  // MUST be field 0
+///          // ... application fields follow
+///      }
```
fbaeuerle, line 33: Shouldn't the file_identifier should also be documented here? What happens if the user doesn't add it?

Would it be possible to serialize? And what happens on calls of GetVersion/VerifyVersion? The function used underneath is undefined in that case: https://github.com/google/flatbuffers/blob/master/include/flatbuffers/buffer.h#L182

Are there enough guards in place to protect against this?

fbaeuerle (changes requested):
i'd like to get the question about file_identifier clarified, the others are rather for my understanding or minor.

examples/integration/flatbuffers/config_usecase/demo_versioned.fbs
OliverHeilwagen: No, i missed to update the comment. The identifiers were different at some point, but i decided to keep them identical for the demos.

score/flatbuffers/common/buffer_version.fbs
OliverHeilwagen: Good catch. I will add this test case and extend the documentation.
serialization is possible without file identifier.

I will check how to guard this case.

score/flatbuffers/common/buffer_version.fbs
OliverHeilwagen: No, i dropped the required to enable the coverage to reach 100%, otherwise type_marker cannot be null as the builder will reject this case. I may change this once we transition to llvm-cov and have generic annotation.
For this PR i would only fix the comment.

score/flatbuffers/common/buffer_version.fbs
OliverHeilwagen: Locally i now got a view on the topic.
The vtable is present in this area if the file identifier is missing (depending on padding).

I will introduce a readable character sanity check this can detect schemas without file_identifier and root table fields < 8214 fields. So in practise it should be robust but there are theoretical schemas which can bypass this guard.

If the number of fields in the root table are >= 8214 the GetVersion can get fooled depending on the padding a "valid" BufferVersionInfo is then returned, bu for the guard use case this should be sufficient.

I need to do some cleanup in the UT, hopefully i can push on Monday.

Update:
In case of deduplicated root vtable (recursive schema and padding alignment), the guard can also be tricked as it will read the negative vtable offset (start of table).

anmittag:
@OliverHeilwagen changes are requested for your PR. are you planning to respond or delete the PR?

OliverHeilwagen:
@anmittag The review findings from @fbaeuerle should be resolved with commit: https://github.com/eclipse-score/baselibs/pull/174/changes/f3aa71ba5ce7c68f44949af5dd81275145db4694.
I am currently waiting for re-review.

fbaeuerle (approved):
✔️

anmittag (approved):

4og (approved):

PR #135, opened 2026-04-01
Add round-trip and leap-year coverage tests for DateTimeConverter (pre/post 1970)

Rutuja-Patil-Bosch:
### Summary
This PR extends unit test coverage for DateTimeConverter along with necessary code change by adding round-trip validation and additional leap-year scenarios, especially around pre-1970 and post-1970 boundaries.

### Changes
- Updated code to get correct values for date & time.
- Added round-trip test validation:
  - epoch → DateTime → epoch consistency checks
- Added test coverage for:
  - Pre-1970 leap year (1968)
  - Post-1970 leap year (1972)
- Included full month/day sweeps to validate:
  - Leap day handling (Feb 29)
  - Month boundary correctness
  - Internal day-sum adjustments

### Motivation
The added tests ensure correctness of:
- Leap year handling logic
- Edge cases around Unix epoch (1970)
- Internal date-to-epoch conversion consistency

This improves confidence in DateTimeConverter for:
- Historical timestamps (pre-1970)
- Leap year transitions
- Boundary conditions in conversion logic

### Notes
- Invalid dates are skipped during iteration
- Focus is on ensuring no regression in conversion logic
- No production code changes included

### Test Scope
- Covers both:
  - 'dateTimeToEpoch'
  - 'epochToDateTime'
- Validates bidirectional consistency

### Impact
- Increased unit test coverage
- Better validation of edge cases and leap-year logic
- No impact on runtime behavior

PR #185, opened 2026-05-08
Bgsw baselib ut reviewfinding fix

Rutuja-Patil-Bosch:
Review findings are fixed for
1. The yearIsLeap() function expects a uint16_t argument, but you're passing an int. When strict warnings are enabled, this type mismatch is treated as a compilation error.
    -> This logic is removed & test cases are updated with fixed input values instead of for loop
2. "Negative" isn't one of the applicable values https://eclipse-score.github.io/process_description/main/process_areas/verification/guidance/verification_specification.html#test-specification
   -> as per spec & discussion in followed comments updated as "requirements-based"
3. Those properties are not part of the test specification https://eclipse-score.github.io/process_description/main/process_areas/verification/guidance/verification_specification.html#test-specification
   -> Removed these additional properties
4. this is dead code
   -> Test cases are updated as dead code was not for all the conditions. Due to for loop, it was was dead code for few iteration. Removed for loop & tested for exact values.


Rutuja-Patil-Bosch:
I'll create this as part of #135

PR #192, opened 2026-05-13
result library fully nodiscard

ssingh2099:
Make result fully nodiscard to prevent ignored return values.

### DIFF
--- a/score/result/details/expected/expected.h
+++ b/score/result/details/expected/expected.h
@@ -76,7 +76,7 @@ template <typename E>
 // copy assignment operators.
 // NOLINTBEGIN(cppcoreguidelines-special-member-functions): shall be implicitly declared as per the C++23 standard.
 // coverity[autosar_cpp14_a12_0_1_violation]
-class unexpected : impl::base_unexpected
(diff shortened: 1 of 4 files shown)


score/result/details/expected/expected.h
```diff
@@ -76,7 +76,7 @@ template <typename E>
 // copy assignment operators.
 // NOLINTBEGIN(cppcoreguidelines-special-member-functions): shall be implicitly declared as per the C++23 standard.
 // coverity[autosar_cpp14_a12_0_1_violation]
-class unexpected : impl::base_unexpected
+class [[nodiscard]] unexpected : impl::base_unexpected
```
fbaeuerle, line 79: is this intentionally made nodiscard on class-level?
it's a divergence from std::unexpected https://en.cppreference.com/cpp/utility/expected/unexpected
If it is deliberate, i would state the reason for it.

fbaeuerle: for unexpected it doesn't have much of an effect anyway (contrary to expected), so it should be fine.

fbaeuerle (approved):

4og (approved):

PR #197, opened 2026-05-15
Build ACL from source

nradakovic:
The PR provides patch for building ACL from release.


nradakovic:
@4og Here is the temp solution until we do not get dedicated modules for ACL and ATTR. As you can see I had to change ACL wrapper where I pass now nullptr (not waiting acl_get_file("") to produce an invalid ACL handle). If the solution is not good, feel free to provide better one.

PR #204, opened 2026-05-20
Add junit xml upload

FScholPer:
Add a junit xml upload of all tests which were executed to later on test which tests do have requirements linkage

Frank Scholter Peres <frank.scholter_peres@mercedes-benz.com>, Mercedes-Benz Tech Innovation GmbH
[Provider Information](https://github.com/mercedes-benz/foss/blob/master/PROVIDER_INFORMATION.md)

### DIFF
--- a/.github/workflows/build_linux.yml
+++ b/.github/workflows/build_linux.yml
@@ -73,6 +73,13 @@ jobs:
         run: |
           # Test that baselibs works when consumed as an external module dependency
           cd examples/integration && bazel build //... && bazel mod deps --lockfile_mode=update
+      - name: Upload JUnit XML test reports
+        if: always()
+        uses: actions/upload-artifact@v7
+        with:
+          name: bazel-testxml-${{ matrix.bazel-config }}
+          path: bazel-testlogs/**/test.xml
+          if-no-files-found: warn
       - name: Upload test logs on failure
         if: failure()
         uses: actions/upload-artifact@v7
--- a/.github/workflows/build_linux.yml
+++ b/.github/workflows/build_linux.yml
@@ -75,7 +75,7 @@ jobs:
           cd examples/integration && bazel build //... && bazel mod deps --lockfile_mode=update
       - name: Upload JUnit XML test reports
         if: always()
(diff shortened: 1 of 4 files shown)


4og:
For this to work, the docs generation job need to be in the same workflow with the build jobs and the jobs should depend on each other. Or the docs workflow uses the `workflow_run` trigger. Here is an example: https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#using-data-from-the-triggering-workflow

I'm not sure what approach is better. I need to think about it.

FScholPer:
hm I will check and make a proposal

MaximilianSoerenPollak:
You probably want to also change the folder you upload this to. 
Take a look at the Docs-As-Code approach that uploads the tests to a `tests-report` folder. That one is also scanned by the source_code_linker, and has not a chance to get overwritten or something by bazel. 

[Here the Link to it](https://github.com/eclipse-score/docs-as-code/blob/main/.github/workflows/test.yml#L46-L60) 

And yes, if you want the links displayed in the documentation it needs to be run in the same workflow. 
[Here an example](https://github.com/eclipse-score/docs-as-code/blob/main/.github/workflows/test_and_docs.yml#L42-L60)

We basically let the tests from another Workflow run in the previous step and then consume the uploaded artifact in the next step.

4og:
@MaximilianSoerenPollak Does docs-as-code support multiple input test reports (for example from QNX and from Linux jobs)? In such case one test may appear several times.

MaximilianSoerenPollak:
> @MaximilianSoerenPollak Does docs-as-code support multiple input test reports (for example from QNX and from Linux jobs)? In such case one test may appear several times.

Each testcase in each test.xml that you provide it will be parsed and created. 
So if there are multiple ones (one QNX & one from Linux) it will be create one for each. 
I would say this is intended behaviour as you want to have knowledge of both executions?

If this is not wanted one solution would be to only copy one of the tests into the folder that is parsed.

4og:
Already done, closing

PR #229, opened 2026-06-03
Add InitializeRuntime API to logging

Y-Vaishnavi:
- Added [InitializeRuntime] as a new public API in [logging.h]
- Implemented in [logging.cpp] using [score::cpp::ignore] to trigger [GetRecorder] and [GetFallbackRecorder]
- No public API existed to initialize the logging runtime, forcing consumers to reach into detail::Runtime directly 
- This provides a proper public alternative that is safe to call multiple times


4og:
This code is already available upstream

Issue #202, opened 2026-05-19
Add test suite for vajson parsing functionality


4og:
### What

Add test suite for the newly integrated vajson library.

### Acceptance Criteria (DoD)

- [x] A test suite implemented for vajson parsing
- [x] Tests cover a variety of valid JSON structures and error cases

PR #237, opened 2026-06-05
Vajson: Add unit tests

JuriSchroeder:
Treats #202. @4og Converting the vaJson unit tests went faster than expected. I kept the original file structure because the repo becomes a bit cluttered when placing the files next to their implementation. Let me know if I should change it for consistencies sake. I also removed some dead code from the implementation.


score/json/internal/parser/vajson/BUILD
```diff
@@ -68,11 +68,48 @@ cc_test(
     ],
 )
 
+cc_test(
```
fbaeuerle: I suggest to tear this build file and this large test target apart into separate build files with separate targets. That's the idiomatic way to do this with bazel (i.e. one BUILD file  per directory with small grained targets). Bazel then has better control over it's cache (e.g. if a parser test is modified, only the parser unit test suite has to be rebuilt).

fbaeuerle: And to add, i would also suggest to use a glob. It can easily happen that someone adds a new test and forgets to explicitly list it here. Thus it would never get built/executed.

fbaeuerle (changes requested):

score/json/internal/parser/vajson/BUILD
```diff
@@ -73,6 +73,7 @@ cc_unit_test_suites_for_host_and_qnx(
     cc_unit_tests = [
         ":vajson_parser_test",
         ":number_parser_test",
+        "//score/json/internal/parser/vajson/unit_test:vajson_impl_unit_tests",
```
4og: This is not really needed. These test suites have no effect on test execution in baselibs (we'll need to clean them up in the whole repo at some point).

4og (changes requested):
The tests are not executed in CI. Can you please remove `target_compatible_with` from `score/json/internal/parser/vajson/BUILD`? It's not needed anymore as we have complete vajson parser implementation.

4og:
The branch needs to be updated. Some jobs are failing because of it.

fbaeuerle (approved):

4og (approved):

PR #261, opened 2026-06-10
Baselibs AutoSD Fixes

odra:
## Changes

* Fixes warnings for GCC 14 (used in AutoSD)
* ~Explicit usage of `-D_GNU_SOURCE`  for GCC 14 (does not break GCC 12)~

### DIFF
--- a/score/concurrency/future/base_interruptible_future_test.cpp
+++ b/score/concurrency/future/base_interruptible_future_test.cpp
@@ -316,7 +316,7 @@ TYPED_TEST(BaseInterruptibleFutureTest, WaitReturnsOnlyAfterValueWasSet)
 
     BaseInterruptibleFutureTest<TypeParam>::SetPromise(this->promise_);
 
-    auto expected = std::move(async_future.get());
+    auto expected = async_future.get();
     EXPECT_TRUE(expected.has_value());
 }
 
@@ -351,7 +351,7 @@ TYPED_TEST(BaseInterruptibleFutureTest, WaitForReturnsTimeoutErrorWhenValueWasNo
         return this->future_->WaitFor(stop_token, TIMEOUT);
     });
 
-    auto expected = std::move(async_future.get());
+    auto expected = async_future.get();
     ASSERT_FALSE(expected.has_value());
     EXPECT_EQ(expected.error(), Error::kTimeout);
 }
@@ -388,7 +388,7 @@ TYPED_TEST(BaseInterruptibleFutureTest, WaitForReturnsWhenValueIsSet)
     });
     BaseInterruptibleFutureTest<TypeParam>::SetPromise(this->promise_);
 
-    auto expected = std::move(async_future.get());
+    auto expected = async_future.get();
     EXPECT_TRUE(expected.has_value());
 }
 
@@ -400,7 +400,7 @@ TYPED_TEST(BaseInterruptibleFutureTest, WaitUntilReturnsTimeoutErrorWhenValueWas
         return this->future_->WaitUntil(stop_token, std::chrono::steady_clock::now() + TIMEOUT);
     });
 
-    auto expected = std::move(async_future.get());
+    auto expected = async_future.get();
     ASSERT_FALSE(expected.has_value());
     EXPECT_EQ(expected.error(), Error::kTimeout);
 }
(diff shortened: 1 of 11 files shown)


4og:
@fbaeuerle, can you please have a look?

third_party/openssl/openssl_transition.bzl
```diff
@@ -39,19 +39,21 @@ _FEATURES_TO_DISABLE = [
 
 def _openssl_transition_impl(settings, attr):
     current = list(settings["//command_line_option:features"])
+    current_copts = settings.get("//command_line_option:copt", [])
```
4og: It's now possible to inject an own openssl lib via a `label_flag`:

https://github.com/eclipse-score/baselibs/blob/41b4af6b955ab550809558233e82e4df609ad8d5/third_party/openssl/BUILD#L17-L25

Would this be a better option than changing compile options in this transition?

odra: Is the idea here to provide an alternative Bazel module / component to manage the openssl build / integration in baselibs?

4og: Correct, it gives an option to substitute openssl from BCR with the openssl provided by the system.

odra: ok, that makes sense. let me remove my openssl change from this pr then.

score/concurrency/future/base_interruptible_future_test.cpp
paulquiring, line 1: Looks like redundant-move was issued as warning, to remove the redundant moves is legitimate.

score/containers/dynamic_array_test.cpp
```diff
@@ -263,8 +263,9 @@ TYPED_TEST(DynamicArrayTestFixture, SelfMoveAssign)
     DynamicArray<TrivialType, TypeParam> unit{kNonEmptyArraySize,
                                               GetAllocator<TrivialType, TypeParam>(this->memory_resource_)};
 
-    // when doing a self-move-assign
-    unit = std::move(unit);
+    // when doing a self-move-assign (use pointer indirection to avoid compiler warning)
```
paulquiring, line 266: The self move warnings can be suppressed like this

```c+
    DISABLE_WARNING_PUSH
    DISABLE_WARNING_SELF_MOVE

    scope = std::move(scope);
```
[see ](https://github.com/eclipse-score/baselibs/blob/d0145e0645d4fe9661a36c348f0e62e13cf73f98/score/language/safecpp/scoped_function/scope_test.cpp#L126)

score/json/internal/parser/parsers_test_suite.h
```diff
@@ -195,7 +195,11 @@ TYPED_TEST_P(ParserTest, CanParseObjectInObject)
     auto root = TypeParam::FromBuffer(buffer_simple_json);
 
     // When reading a key of an object that is interpreted as number
-    auto& value = GetValueOfObject<Object>(root.value(), "object");
```
paulquiring, line 198: I wonder if just removing the & would also work to get rid of the dangling-reference error. Above it seems to work.

paulquiring (changes requested):

score/json/internal/parser/parsers_test_suite.h
odra: It fixed it for primitive types but it required a little bit for objects and lists.

score/language/safecpp/scoped_function/details/allocator_aware_type_erasure_pointer_test.cpp
```diff
@@ -272,7 +272,8 @@ TEST_F(AllocatorAwareTypeErasurePointerTest, CanMoveAssignSelfWithoutAdverseEffe
     DISABLE_WARNING_PUSH
     DISABLE_WARNING_SELF_MOVE
 
-    target = std::move(target);
+    auto* target_ptr = &target;
+    target = std::move(*target_ptr);
```
4og, line 276: How does the compiler warning look like?
With the `DISABLE_WARNING_PUSH` above there shouldn't be the need to change this code.

score/containers/dynamic_array_test.cpp
odra: I though it would made more sense to just fix the warnings instead of just ignoring it.

paulquiring (commented):
some findings

4og:
There is an ongoing work to setup the GCC 15 toolchain in S-CORE: https://github.com/eclipse-score/bazel_cpp_toolchains/issues/100
In my local testing I see the issues this PR is trying to address. I hope soon we'll also have GCC15 in the CI.

odra:
@4og since it's going to be addressed once gcc15 is enabled in S-CORE's CI pipelines, should we close this PR then?

PR #263, opened 2026-06-10
score/containers: Extend test cases

crimson11:
Unit test have been extended to also verify
begin/end pointer dereferencing to support
bounds-checking.


crimson11:
needed to push from fork.

Issue #203, opened 2026-05-19
Add AoUs, safety analysis, detailed design for vajson


4og:
### What

Add the following documentation pages for vajson: assumptions of use, safety analysis, detailed design.

### Acceptance Criteria (DoD)

- [ ] The documentation is added in the S-CORE format according to https://eclipse-score.github.io/process_description/main/folder_templates/modules/module_name/component_name/docs/index.html Alternatively, the documentation can be added in any other format (there will be follow up tasks to convert).

4og:
Still in progress

PR #298, opened 2026-06-24
Add vaJson detailed design

JuriSchroeder:
This PR dumps the vaJson detailed design as requested by #203 . It has not been adapted to the S-CORE format yet, but I adapted the namespaces / package names and removed those parts that are not applicable to S-CORE (binary parsing, dependencies to Vector internal libraries, etc.)
Safety artifacts will be dumped in a separate MR.


4og (approved):
This looks good to me.
@aschemmel-tech, if you have any concerns about the structure, let's have a fix it as a follow up

PR #299, opened 2026-06-24
Integrate clang-format via the new score_tooling format_checker

4og:
Switch C++ formatting to the configurable `use_format_targets` API introduced in eclipse-score/tooling#294, and stop using the old `--config clang_format` aspect path.

- Enable `cpp` in `use_format_targets(languages = [...])`, which wires `clang-format` into `//:format.fix` and `//:format.check`.
- Keep the `languages = [...]` support local for now via a small wrapper macro, instead of pinning `score_tooling` to the PR-294 commit. This avoids pulling in unrelated newer transitive Bazel module dependencies during CI.
- Drop the old wiring: the `clang-format` CI job (`bazel build --config clang_format`), the `build:clang_format` block in .bazelrc, and the `clang_format_config` target in BUILD. The reusable `formatting-check` job already runs `bazel test //:format.check`, which now covers C++.
- Keep `.clang-format-ignore` so currently-unformatted folders stay excluded for now.


4og:
@copilot Why do CI jobs fail when attempting to use rules_android? This bazel module doesn't use anything Android related.

.clang-format-ignore
```diff
@@ -0,0 +1,91 @@
+# Limit clang-format enforcement for now. Each block below ignores one folder's
```
fbaeuerle, line 1: seems to me like we _could_ align this with the [restricted paths](https://github.com/eclipse-score/baselibs/blob/main/.github/workflows/restricted_paths.yml#L37)

do you have reasons against it?

4og: The paths lists are not exactly the same as we have libs removed from the restricted paths, but still not reformatted. My goal that the `.clang-format-ignore` becomes empty eventually once the restrictions are lifted and the code is reformatted.

Moreover, It's hard to come up with a common list as the file path matching syntax in ignore file and in `dorny/paths-filter` is not the same.

antonkri (approved):

PR #318, opened 2026-06-29
score/containers: Extend test cases

crimson11:
Unit test have been extended to also verify
begin/end pointer dereferencing to support
bounds-checking.


score/containers/non_relocatable_vector_test.cpp
```diff
@@ -237,4 +240,136 @@ TYPED_TEST(NonRelocatableVectorNonMoveableAndCopyableElementTypeFixture, SwapSwa
     }
 }
 
+class NonRelocatableVectorPointerSpyFixture : public ::testing::Test
+{
+    void SetUp() override {}
+    void TearDown() override {}
```
fbaeuerle: this seems unnecessary

score/containers/non_relocatable_vector_test.cpp
```diff
@@ -237,4 +240,136 @@ TYPED_TEST(NonRelocatableVectorNonMoveableAndCopyableElementTypeFixture, SwapSwa
...
+    {
+        unit_ = std::make_unique<NonRelocatableVector<ElementType, Allocator>>(number_of_elements);
+        for (std::size_t i = 0; i < number_of_elements; ++i)
+        {
+            score::cpp::ignore = this->unit_->emplace_back(i);
+        }
+        return *this;
+    }
+
+    score::memory::shared::test::MyMemoryResource memory_resource_{};
```
fbaeuerle: is this unused or am i overseeing something?

score/containers/non_relocatable_vector_test.cpp
```diff
@@ -237,4 +240,136 @@ TYPED_TEST(NonRelocatableVectorNonMoveableAndCopyableElementTypeFixture, SwapSwa
...
+    std::unique_ptr<NonRelocatableVector<ElementType, Allocator>> unit_{nullptr};
+};
+
+using NonRelocatableVectorPointerInteractionFixture = NonRelocatableVectorPointerSpyFixture;
```
fbaeuerle: can we spare this alias and call the fixture NonRelocatableVectorPointerInteractionFixture directly?

score/containers/non_relocatable_vector_test.cpp
```diff
@@ -237,4 +240,136 @@ TYPED_TEST(NonRelocatableVectorNonMoveableAndCopyableElementTypeFixture, SwapSwa
...
+
+        // expect, that the pointer to the start of the data is advanced to the last element (size - 1) ...
+        EXPECT_CALL(pointer_spy, PlusAssign(static_cast<std::ptrdiff_t>(kNonZeroNumberElements - 1)));
+        // ... and two pointer dereferences take place
+        EXPECT_CALL(pointer_spy, Arrow(_)).Times(2).WillRepeatedly(testing::Invoke([&arrow_args](ElementType* ptr) {
+            // capturing which addresses are dereferenced.
+            arrow_args.push_back(ptr);
+        }));
+
+        // when calling data() on the NonRelocatableVector
```
fbaeuerle: this complete part is duplicated 3 times (except for the variable name).

The mock setup could be outsourced to a helper function that takes a lambda (for the When-part) and returns the variables required for the Then-part.

fbaeuerle (changes requested):

score/containers/non_relocatable_vector_test.cpp
crimson11: you are correct! Removed.

score/containers/non_relocatable_vector_test.cpp
crimson11: yep. Makes sense as we aren't using the `NonRelocatableVectorPointerSpyFixture` in another/semantical different setup.

score/containers/non_relocatable_vector_test.cpp
crimson11: Good idea! Will try to compact/reduce duplication.

fbaeuerle (approved):
👍

score/containers/non_relocatable_vector_test.cpp
fbaeuerle: looks much nicer now, thanks!

4og (approved):

PR #332, opened 2026-07-01
tooling: improve IDE support

arkjedrz:
- Add VS Code configuration.
  - Rust formatting, IDE support, basic tasks.
- Add `compile_commands.json` and `rust-project.json` generation.
- Patch to `hedron_compile_commands` for Bazel 8.6.0 support.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,4 +1,24 @@
 {
+    "[rust]": {
+        "editor.defaultFormatter": "rust-lang.rust-analyzer",
+        "editor.rulers": [
+            150
+        ]
+    },
+    "rust-analyzer.cargo.cfgs": [],
+    "rust-analyzer.cargo.features": [],
+    "rust-analyzer.cargo.noDefaultFeatures": false,
+    "rust-analyzer.check.command": "clippy",
+    "rust-analyzer.rustfmt.overrideCommand": [
+        "${workspaceFolder}/.vscode/rustfmt.sh"
+    ],
+    "clangd.arguments": [
+        "--header-insertion=never",
+        "--compile-commands-dir=${workspaceFolder}/",
+        "--query-driver=**",
+        "--clang-tidy",
+        "--fallback-style=None"
+    ],
     "json.schemas": [
         {
             "fileMatch": [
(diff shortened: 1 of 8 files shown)


pawelrutkaq (dismissed):
@4og can you look ?

.vscode/settings.json
```diff
@@ -1,4 +1,33 @@
 {
+    "[rust]": {
+        "editor.defaultFormatter": "rust-lang.rust-analyzer",
+        "editor.formatOnSave": true,
+        "editor.rulers": [
+            150
+        ]
+    },
+    "rust-analyzer.cargo.cfgs": [],
```
4og, line 8: If found this page: https://bazelbuild.github.io/rules_rust/rust_analyzer.html
Wouldn't such approach better for users?

arkjedrz:
I checked what will happen when plugin is disabled. Plugin-specific options are grayed out - this is fine.
`defaultFormatter` fields raise a warning that value is not accepted - this is non-breaking for the most part. If a person wants to use a different tooling for VS Code then `settings.json` can be modified locally. This config provides simple tooling setup and reproducible formatting for most.

obraz

.vscode/settings.json
```diff
@@ -1,4 +1,30 @@
...
+        "--query-driver=**",
+        "--clang-tidy",
+        "--fallback-style=None"
+    ],
+    "[cpp]": {
+        "editor.defaultFormatter": "llvm-vs-code-extensions.vscode-clangd",
+    },
+    "[c]": {
+        "editor.defaultFormatter": "llvm-vs-code-extensions.vscode-clangd",
+    },
```
4og: I prefer not to have it. The default C++ extension from vscode supports clang-format out of the box without extra configuration. Having this her, will break the formatter for those who don't use vscode-clangd extension

4og (changes requested):

.vscode/settings.json
arkjedrz: We're using `rules_rust` fork based on v0.68.1, this feature is not there yet.

4og (approved):

PR #336, opened 2026-07-02
os: Change glibc < 2.30 warn to pragma message

Ivan-Tiurin:
- To not block building with old glibc and -Werror


fbaeuerle (approved):
This allows compilation with -Werror -Wpedantic with toolchains that have glibc < 2.30.

4og (approved):

PR #342, opened 2026-07-06
OS: Add QNX I/O and OCB wrappers

pushkaraj-gode:
- Add wrappers for io_close_dup_default, io_lock_ocb_default and io_unlock_ocb_default. 
- Added corresponding mocks
- Unit tests to improve coverage.

### DIFF
--- a/score/os/mocklib/qnx/mock_iofunc.h
+++ b/score/os/mocklib/qnx/mock_iofunc.h
@@ -138,6 +138,21 @@ class MockIoFunc : public IoFunc
                 iofunc_notify_remove,
                 (resmgr_context_t * ctp, iofunc_notify_t* nop),
                 (const, noexcept, override));
+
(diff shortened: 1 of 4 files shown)


Ivan-Tiurin (approved):
**LGTM**. Using raw std::int32_t return for these callback-oriented wrappers (iofunc_close_dup_default, iofunc_lock_ocb_default, iofunc_unlock_ocb_default) is reasonable wrapping in expected_blank only to unwrap at the callback site adds overhead with no real benefit.

**For future** consideration: the existing expected_blank<std::int32_t> pattern in sibling functions (e.g. iofunc_attr_lock) has the same issue - a raw int as the error type isn't more descriptive than the return code itself. A proper Error enum (like iofunc_mount_init uses) would justify expected_blank; without one, it's just int-to-int ceremony. Worth a follow-up to align the module.

4og (approved):

PR #369, opened 2026-07-13
os: justification for the AoU

Marcin-GA-Grzelak:


Marcin-GA-Grzelak:
not required

> There is no list of banned methods. We don't need argumentation for lib/os as it will not appear in clang findings on SPP side, and we know for sure that baselibs/os contains wrappers so it's safe. We checking use of score::os::mmap on SPP side

PR #370, opened 2026-07-13
refactor: add SCORE_ prefix to static reflection macros

rahulthakre29:
- prefix static reflection macros with SCORE_
- update generator and tests
- fix tests to use the new macro names


4og:
Please also update markdown documentation

rahulthakre29:
> Please also update markdown documentation

Done

score/static_reflection_with_serialization/visitor/include/visitor/visit_as_struct.h
```diff
@@ -230,619 +230,629 @@ using struct_visitable = decltype(get_struct_visitable<T>());
...
+
+// Deprecated aliases kept for backward compatibility.
+// Use SCORE_STRUCT_VISITABLE and SCORE_STRUCT_TRACEABLE in new code.
+#ifndef STRUCT_VISITABLE
+#define STRUCT_VISITABLE(...) SCORE_STRUCT_VISITABLE(__VA_ARGS__)
+#endif
+
+#ifndef STRUCT_TRACEABLE
+#define STRUCT_TRACEABLE(...) SCORE_STRUCT_TRACEABLE(__VA_ARGS__)
+#endif
```
4og, line 861: It would be nice to introduce some way to emit compiler warnings when a deprecated macro is used.

Maybe something like this?
```cpp
#define SCORE_DEPRECATE_MACRO_USE(MESSAGE) SCORE_DEPRECATE_MACRO_USE_IMPL(MESSAGE, __COUNTER__)
#define SCORE_DEPRECATE_MACRO_USE_IMPL(MESSAGE, ID)                                                      \
    using score_deprecated_macro_def_##ID [[deprecated(MESSAGE)]] = void;                                \
    using score_deprecated_macro_use_##ID [[maybe_unused]] = score_deprecated_macro_def_##ID
```

and usage:

```cpp
// Deprecated aliases kept for backward compatibility.
// Use SCORE_STRUCT_VISITABLE and SCORE_STRUCT_TRACEABLE in new code.
#ifndef STRUCT_VISITABLE
#define STRUCT_VISITABLE(...)                                                                                         \
    SCORE_DEPRECATE_MACRO_USE("STRUCT_VISITABLE is deprecated, use SCORE_STRUCT_VISITABLE instead.");                 \
    SCORE_STRUCT_VISITABLE(__VA_ARGS__)
#endif

#ifndef STRUCT_TRACEABLE
#define STRUCT_TRACEABLE(...)                                                                                         \
    SCORE_DEPRECATE_MACRO_USE("STRUCT_TRACEABLE is deprecated, use SCORE_STRUCT_TRACEABLE instead.");                 \
    SCORE_STRUCT_TRACEABLE(__VA_ARGS__)
#endif
```

I'd put the `SCORE_DEPRECATE_MACRO_USE` to
https://github.com/eclipse-score/baselibs/blob/main/score/quality/compiler_warnings/warnings.h

rmaddikery: I would keep it in visit_as_struct.h where it is currently used - better for tracking and removal later. warnings.h serves other purpose and better not polluted.

Another sidenote:
https://en.cppreference.com/cpp/language/attributes/deprecated
[[deprecated]] applies to aliases and not macros. So double check.

score/quality/compiler_warnings/warnings.h
```diff
@@ -77,3 +77,13 @@
 #define DISABLE_WARNING_MISSING_DESIGNATED_FIELD_INITIALIZERS
 
 #endif
+
+// clang-format off
+// Emit deprecation warnings from macro aliases by referencing a deprecated type alias.
+#ifndef SCORE_DEPRECATE_MACRO_USE
+#define SCORE_DEPRECATE_MACRO_USE(MESSAGE) SCORE_DEPRECATE_MACRO_USE_IMPL(MESSAGE, __COUNTER__)
+#define SCORE_DEPRECATE_MACRO_USE_IMPL(MESSAGE, ID)                                                      \
+    using score_deprecated_macro_def_##ID [[deprecated(MESSAGE)]] = void;                               \
```
rmaddikery: Was curious how this works, I see gcc fires the warning twice! Here's a trick to fire it once:

```suggestion
    using score_deprecated_macro_def_##ID [[deprecated(MESSAGE)]] = int;
    static_assert(sizeof(score_deprecated_macro_def_##ID) > 0, "")                               \
```

score/quality/compiler_warnings/warnings.h
```diff
@@ -77,3 +77,13 @@
 #define DISABLE_WARNING_MISSING_DESIGNATED_FIELD_INITIALIZERS
 
 #endif
+
+// clang-format off
+// Emit deprecation warnings from macro aliases by referencing a deprecated type alias.
+#ifndef SCORE_DEPRECATE_MACRO_USE
```
rmaddikery: As said, its better to seperate the concers. The warnings.h is solelz for suppressions. We  are trying to emit warnings. Keep them seperate. In this case move it to the source.

score/quality/compiler_warnings/warnings.h
rahulthakre29: Done

score/quality/compiler_warnings/warnings.h
rahulthakre29: Done

rmaddikery (approved):

4og (approved):

limdor:
This introduced a deprecation warning to everyone even if they use the none deprecated macros.
The reason for that is because of the include https://github.com/eclipse-score/baselibs/blob/36ace4168b03c2aa1b14133d0da7078e3320247b/score/static_reflection_with_serialization/visitor/include/visitor/visit_as_struct.h#L23

limdor:
> This introduced a deprecation warning to everyone even if they use the none deprecated macros. The reason for that is because of the include
> 
> https://github.com/eclipse-score/baselibs/blob/36ace4168b03c2aa1b14133d0da7078e3320247b/score/static_reflection_with_serialization/visitor/include/visitor/visit_as_struct.h#L23

It seems it came from mw/log still using the old macro
With https://github.com/eclipse-score/baselibs/pull/442 the new warnings I saw are going away.

PR #377, opened 2026-07-14
const overloads & nodiscard for LockedPtr members

sankurm:
This prepares the `LockedPtr` for the monadic interface that it should have. The support functions
1. make the return of functions mandatory to handle `[[nodiscard]]`
2. Ensure that `const` overloads are supported (needed for further upcoming functionality)


fbaeuerle (commented):
i'd like to see copilots suggestion implemented (i.e. testing the behavior also for the const version)

fbaeuerle (approved):

4og (approved):

Issue #248, opened 2026-06-08
Relocate Rust source code from `src` into component folders under `score`


4og:
## What

As part of the `baselibs_rust` migration (follow-up to #200), the Rust source code was initially placed under a flat src folder. This is inconsistent with the existing `baselibs` structure, where each component lives under `score/<component>/`. Relocating the Rust modules into the appropriate score subtrees aligns the repository layout, makes cross-language components discoverable in one place, and enables unified Bazel visibility and documentation rules.

## Acceptance Criteria (DoD)

- [ ] An agreed target location under score is defined for each module currently in src: `containers`, `log` (comprising `score_log`, `score_log_fmt`, `score_log_fmt_macro`, `stdout_logger`, `stdout_logger_cpp_init`), `pal`, `thread`, `sync`, `elementary`, and `testing_macros`.
- [ ] All modules are moved from `/src` to their agreed target locations under `/score`.
- [ ] CI builds and tests pass for all relocated Rust targets after the move.
- [ ] The docs references to component source paths are updated to match the new locations.

arkjedrz:
Approach must be discussed. One must be chosen, but none are really convenient for all cases.

Keep in mind that target names match the directory structure in the repository.

### `//score/cpp/<XYZ>` and `//score/rust/<XYZ>`

Having languages split at `//score` level makes sense when features offered for C++ and Rust are not related to each other.

#### Libraries that make sense only for one language

Example: `futurecpp`.

```plain
//score/language/futurecpp --> //score/cpp/language/futurecpp
```

#### Unrelated libraries of the same name

Example: `containers`. Those libraries are not related, only name is shared.

```plain
//score/containers:dynamic_array --> //score/cpp/containers:dynamic_array
//src/containers --> //score/rust/containers
```

#### Mixed libraries

There are also mixed libs, and this would make implementations far away from each other.
E.g., `flatbuffers` are expected to share implementation later on, target names would be:

```plain
//score/flatbuffers:flatbufferscpp --> //score/cpp/flatbuffers
N/A --> //score/rust/flatbuffers
```

This would make Rust interface to be elsewhere in the repository than main C++ lib.

### `//score/<XYZ>/cpp` and `//score/<XYZ>/rust`

This makes mixed libraries code closer to each other.

#### Libraries that make sense only for one language

```plain
//score/language/futurecpp --> //score/language/cpp:futurecpp
```

or just don't care for such scenarios and leave them in the `//score/`:

```plain
[...]

PR #390, opened 2026-07-16
move content of log_rust to /score/mw/log/rust + adoptions due to process

arsibo:
- move /score/log_rust/* to /score/mw/log/rust/ + adaptions + alias setting (as discussed in https://github.com/eclipse-score/baselibs/issues/248)
- add docs folder structure to /score/mw/log/docs (https://eclipse-score.github.io/score/main/contribute/general/folder.html)
- rename of /score/mw/log/test to /score/mw/log/tests due to https://eclipse-score.github.io/score/main/contribute/general/folder.html  
- due to have a proper include path /score/mw/log/.. the following topics will not be implemented
   - src directory (/score/mw/log/src will not introduced due a proper include path  https://eclipse-score.github.io/score/main/contribute/general/folder.html will not be realized )
   - component name log_frontend as component directory name. We will use "Log frontend" in the title of the Sphinx component Element.


4og:
@arkjedrz, @rmaddikery, can you please have a look?

score/mw/log/configuration/BUILD
```diff
@@ -172,7 +172,7 @@ cc_test(
         "@score_baselibs//score/os/mocklib:stdlib_mock",
         "@score_baselibs//score/os/mocklib:unistd_mock",
         "@score_baselibs//score/os/utils/mocklib:path_mock",
-        #"@score_baselibs//score/mw/log/test/console_logging_environment",
+        #"@score_baselibs//score/mw/log/tests/console_logging_environment",
```
arkjedrz, line 175: IMO those commented-out lines should be removed separately.

arkjedrz (commented):
Overall LGTM, but it was also discussed that `rust`/`cpp` subdirectory should be used when a component offers secondary interface to another language (library X written in C++ provides Rust interface -> `X/rust`, library Y written in Rust provides C++ interface -> `Y/cpp`).
This is not the case here, as Rust log implementation is a separate entity with C++ implementation being used as one of the backends.
Please confirm this is accepted both by logging and baselibs FT.

PR #393, opened 2026-07-17
Add size_bytes() method to span

mariuswbr:
Add size_bytes() method support for span to be compliant with std::span (C++20).

### DIFF
--- a/score/language/futurecpp/include/score/span.hpp
+++ b/score/language/futurecpp/include/score/span.hpp
@@ -291,7 +291,10 @@ public:
 
     /// \brief Returns the size of the row in bytes.
     ///
-    /// \return  Size of the row in bytes.
+    /// \note Counts only `size() * sizeof(T)`. Heap storage owned by allocator-aware element types
+    /// (e.g. `std::vector`) is not included, consistent with `std::span::size_bytes`.
+    ///
+    /// \return Size of the row in bytes.
     size_type size_bytes() const noexcept { return base_.size() * sizeof(T); }
 
     /// \brief checks if the span is empty
--- a/score/language/futurecpp/include/score/span.hpp
+++ b/score/language/futurecpp/include/score/span.hpp
@@ -289,6 +289,11 @@ public:
     /// \return  Number of elements in the row.
     size_type size() const noexcept { return base_.size(); }
 
+    /// \brief Returns the size of the row in bytes.
+    ///
+    /// \return  Size of the row in bytes.
+    size_type size_bytes() const noexcept { return base_.size() * sizeof(T); }
+
     /// \brief checks if the span is empty
     ///
     /// @return true if span is empty (size==0); false otherwise.
--- a/score/language/futurecpp/tests/span_test.cpp
+++ b/score/language/futurecpp/tests/span_test.cpp
@@ -477,6 +477,32 @@ TEST(span, DereferencingNullptrShallTriggerContracViolation)
 
 /// @testmethods TM_REQUIREMENT
 /// @requirement CB-#9338069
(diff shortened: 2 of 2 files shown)


score/language/futurecpp/tests/span_test.cpp
```diff
@@ -475,6 +475,32 @@ TEST(span, DereferencingNullptrShallTriggerContracViolation)
     }
 }
 
+/// @testmethods TM_REQUIREMENT
+/// @requirement CB-#9338069
```
mariuswbr, line 479: Unfortunately I really don't know how this works in baselibs. 
I have seen this was added to all test cases. Let me know if I need to adjust something. 
Thanks!

mariuswbr:
@4og The failing pipeline asks me to contact you :). 
I was trying to add the size_bytes() method to span to be compliant with std::span:
https://en.cppreference.com/cpp/container/span/size_bytes

Is that acceptable?

score/language/futurecpp/tests/span_test.cpp
Ehrja: Currently, the mentioned requirement-numbers are still referring to a BMW requirements-server. This will hopefully change in the mid-future :sweat_smile: 

> Let me know if I need to adjust something.

It should be fine

score/language/futurecpp/tests/span_test.cpp
```diff
@@ -475,6 +475,32 @@ TEST(span, DereferencingNullptrShallTriggerContracViolation)
...
+TEST(span, SizeBytes)
+{
+    {
```
Ehrja, line 482: Would you mind to extend the test-cases by one which specifies an alignment greater as the typical underlying-type would require? 

Maybe something like: 
```cpp
{
    alignas(64) const std::int32_t data[]{23, 42, 72};
    const std::span<const std::int32_t, 3U> view{data};
    EXPECT_EQ(view.size() * sizeof(std::int32_t), view.size_bytes());
}
```

This shall ensure that alignment-requirements does **not** have an influence on the result of `size_bytes`.

score/language/futurecpp/include/score/span.hpp
```diff
@@ -289,6 +289,11 @@ class span
     /// \return  Number of elements in the row.
     size_type size() const noexcept { return base_.size(); }
 
+    /// \brief Returns the size of the row in bytes.
+    ///
```
Ehrja, line 293: I'm thinking about the case when a span is referring to nested containers where the nesting is an allocator aware type and contains some part of its complete footprint on heap. Then the `size_bytes` will actually return less "bytes" than it actually would require. 

Your impl is standard compliant, I'm just thinking about a doxygen `\note` about this behavior 🤔 - what do you think?

Ehrja (commented):
I maintain the BMW internal representation of futurecpp. I would have left the following comments if it would have been an in-house PR.

Please consider them as non-binding since I'm not a score-baselib maintainer :)

score/language/futurecpp/tests/span_test.cpp
mariuswbr: I don't mind at all. Thank you very much for the review.
Added the proposed test case.

score/language/futurecpp/include/score/span.hpp
mariuswbr: I was also thinking about something similiar while having a look at std::span. I like the idea about further clarification and added a \note. Would you mind re-checking if the wording matches your expectation? I tried to keep it short.

Ehrja (approved):
From my perspective it would be ready to merge

4og (approved):

Issue #371, opened 2026-07-13
Relocate `score/memory` to `eclipse-score/communication` and clean up remaining baselibs dependencies


4og:
Relocate the `score/memory` library from `eclipse-score/baselibs` to `eclipse-score/communication`.

## Motivation
- The library fits better to the communication concern in S-CORE.
- It should be maintained by the same team that owns related communication functionality.

## Scope
- Analyze the current contents of `score/memory` and split them into:
  - communication-specific parts that should move to `eclipse-score/communication`
  - generic/base parts that should remain in `eclipse-score/baselibs`
- Clean up dependencies inside `baselibs`, because there are current internal usages of `score/memory`.
- Move the remaining generic pieces into appropriate existing libraries in `baselibs` instead of keeping them in a broad memory package.

## Acceptance criteria
- [x] Communication-specific memory functionality is moved to `eclipse-score/communication`.
- [x] Generic memory utilities still needed by `baselibs` remain there and are placed into suitable existing libraries.
- [x] Internal `baselibs` dependencies on moved components are removed or redirected.

PR #400, opened 2026-07-22
Further preparation for moving score/memory/shared

LittleHuba:
Part of #371

### DIFF
--- a/score/language/safecpp/aborts_upon_exception/BUILD
+++ b/score/language/safecpp/aborts_upon_exception/BUILD
@@ -43,6 +43,8 @@ cc_test(
     ],
 )
 
+# Intentionally fails since toolchain does not yet support this feature
+# See https://github.com/eclipse-score/baselibs/issues/7
 cc_test(
     name = "abortsuponexception_toolchain_test",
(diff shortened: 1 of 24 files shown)


score/language/safecpp/aborts_upon_exception/BUILD
```diff
@@ -52,6 +52,7 @@ cc_test(
     features = ["aborts_upon_exception"],
     tags = ["unit"],
     deps = [
+        ":abortsuponexception",
```
4og: This test is currently failing on purpose as the toolchain in S-CORE doesn't have aborts upon exception activated. We were actually hoping to fix the toolchain (maybe by having an injectable feature, I'm not sure)

4og (changes requested):

score/language/safecpp/aborts_upon_exception/BUILD
LittleHuba: Then we should at least put a comment somewhere. This is really confusing if you run bazel test //...

4og (approved):

PR #402, opened 2026-07-22
Move DataTypeSizeInfo to score/memory/shared

LittleHuba:
Moves the class to score/memory/shared and degrades it to an implementation detail.

score/memory/shared is currently the only user.
This class is dangerous since it gives no guarantee that the
size or alignment are actually correct for the associated type.
Users should prefer a RAII pointer style abstraction instead
(see scoped_function implementation details).

The safest approach is to remove it from the public API and
then clean up the current usages.

Only last commit is relevant. Other commits are from #400 and are required to avoid merge conflicts.


LittleHuba:
Closing since this target is actually used outside of score/memory/shared. Hence the reasoning behind this move is wrong. The move can therefore not be done.

PR #252, opened 2026-06-09
chore(CI): verify bazel lockfile is correct

AlexanderLanin:
Verify correctness of Bazel lockfile on Pull Request

Note: this will fail until lockfile is manually fixed

PR #405, opened 2026-07-23
Add Verify Lockfile Correctness Workflow

paulquiring:
Adds a dedicated workflow that checks lock file correctness


4og:
This looks similar to #252. Why not to use https://github.com/eclipse-score/cicd-workflows/blob/main/.github/workflows/bzlmod-lock-check.yml?

paulquiring:
will be closed with https://github.com/eclipse-score/baselibs/pull/252

PR #407, opened 2026-07-23
Use `grpc-java` dependency instead of `grpc` in the flatbuffers workaround

4og:
The `grpc` version 1.76 requires every baselibs user with bazel <8.6 to place a `single_version_override` for `rules_swift` in their MODULE file to satisfy version compatibility check. We revert back to depending on `grpc-java`.


LittleHuba (approved):
Verified to work in score_communication

PR #409, opened 2026-07-24
analysis/tracing: Use std::variant instead of score::cpp::variant

indramaniarorascore:
This PR changes the generic_trace_api's interfaces and stubs to use std::variant instead of score::cpp::variant.

std::variant is denoted by the type MetaInfoVariants::StdType in meta_info_variants.h

Also, Corresponding helper functions like holds_alternative were changed to the std version.

NOTE: I have checked compiling communication with this change, and this change does NOT affect communication i.e users of generic_trace_api


hoppe-and-dreams (commented):
The content of the PR is fine, BUT
- Please adjust the PR title and the commit message title. Prefix with analysis/tracing so it's clear this is _not_ a repo wide change
- This is a breaking API change, please make this visible. (This is aligned with the only S-CORE user communication but still) 
- Please add a reasonable description to the PR

hoppe-and-dreams (approved):
My comments were addressed. Thanks :)

4og (approved):

PR #415, opened 2026-07-27
analysis/tracing - Remove usage of score::cpp::variant from service_instance_element

indramaniarorascore:
ServiceInstanceElement used element_id of score::cpp::variant type. changed it to std::variant

### DIFF
--- a/score/analysis/tracing/generic_trace_library/interface_types/service_instance_element.h
+++ b/score/analysis/tracing/generic_trace_library/interface_types/service_instance_element.h
@@ -15,7 +15,6 @@
 
 #include <score/utility.hpp>
 #include <score/variant.hpp>
-
 #include <cstdint>
 #include <variant>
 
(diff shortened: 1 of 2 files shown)


score/analysis/tracing/generic_trace_library/interface_types/service_instance_element.h
```diff
@@ -93,7 +85,7 @@ class ServiceInstanceElement
     InstanceIdType instance_id;
     // No harm to declare the members as public
     //  coverity[autosar_cpp14_m11_0_1_violation]
-    VariantType element_id;
```
andreapefe, line 96: As you are removing a VariantType, please check that all dependendees (including internal ones) have migrated to the new version.

andreapefe (changes requested):

score/analysis/tracing/generic_trace_library/interface_types/service_instance_element.h
andreapefe: Checked internally: Migration will be done when upgrading the version not later.

andreapefe (approved):

4og (approved):

PR #420, opened 2026-07-29
Making inotify bazel target public

MaciejKaszynski:
Making the `inotify` bazel target public, (I think it was made non-public by mistake as the mock is public).

### DIFF
--- a/score/os/BUILD
+++ b/score/os/BUILD
@@ -133,9 +133,7 @@ cc_library(
     ],
     features = COMPILER_WARNING_FEATURES,
     tags = ["FFI"],
-    visibility = [
-        "@score_baselibs//score/os:__subpackages__",
-    ],
+    visibility = ["//visibility:public"],
     deps = [
         ":errno",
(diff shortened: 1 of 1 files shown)


score/os/BUILD
```diff
@@ -133,9 +133,7 @@ cc_library(
     ],
     features = COMPILER_WARNING_FEATURES,
     tags = ["FFI"],
-    visibility = [
-        "@score_baselibs//score/os:__subpackages__",
-    ],
+    visibility = ["//visibility:public"],
```
ravindra181, line 136: Why is this visiblity change needed.
Is this target https://github.com/eclipse-score/baselibs/blob/main/score/os/utils/inotify/BUILD not enough ?

ravindra181 (changes requested):

MaciejKaszynski:
A different target shall be used https://github.com/eclipse-score/baselibs/blob/main/score/os/utils/inotify/BUILD

PR #427, opened 2026-07-30
Revert "result library fully nodiscard"

LittleHuba:
Reverts eclipse-score/baselibs#192

This fully breaks public API and is very hard for users to adopt.
@ssingh2099 please come up with a migration path for this. Just dropping this in to let users cope on their own is not an option.


4og (approved):
The nodiscard affects the memory/shared lib migration. This should be brought back once the migration is complete

PR #429, opened 2026-07-30
Fix ASAN support for Rust code

eclipse-impl:


dcalavrezo-qorix:
I have one question regarding this change.

As far as I know, Rust sanitizers are only available with a nightly compiler via:

`-Zsanitizer=address`

and all -Z flags are unstable.

https://rustc-dev-guide.rust-lang.org/sanitizers.html

The changes in this PR appear to enable the experimental rules_rust linking features (experimental_use_cc_common_link and experimental_use_allocator_libraries_with_mangled_symbols), but I don't see where the Rust code itself is actually being ASan-instrumented.

Am I missing something, or does this PR only change the final linking behavior rather than enabling AddressSanitizer for Rust code

eclipse-impl:
> As far as I know, Rust sanitizers are only available with a nightly compiler via:
> 
> `-Zsanitizer=address`

You're right — this PR does not enable `-Zsanitizer=address`, so the Rust code itself is not ASan-instrumented (that needs nightly). What it changes is only the final link: routing it through the C++ toolchain (cc_common.link) so the ASan runtime and its interceptors are actually linked in, and so our sanitized C/C++ code links cleanly with the correct sanitizer recipe instead of us hand-appending archives.

The coverage we get is therefore the interception-based subset of ASan: global malloc/free replacement (so Rust heap allocations get redzones + quarantine), libc interceptors, double-free/invalid-free, and full LSan leak detection — plus complete instrumentation of the C/C++ portions and the FFI boundary. Compile-time-only checks on pure-Rust accesses (stack/global redzones, shadow checks on Rust loads/stores, Rust UBSan) are out of scope until we can adopt -Zsanitizer.

Notice, though, that this brings us much further from where we were, as at least the important subset of ASan that deals with dynamic memory allocations errors is introduced, and already proved its value by finding one bug.

.bazelrc
arkjedrz: Rename section to `Sanitizer configuration for C++ and Rust`.

.bazelrc
```diff
@@ -109,6 +109,18 @@ test:asan_ubsan_lsan --copt -g3
 test:asan_ubsan_lsan --platform_suffix=asan_ubsan_lsan
 test:asan_ubsan_lsan --test_env=ASAN_OPTIONS="exitcode=55 allow_addr2line=1 verbosity=1 coverage=1 check_initialization_order=1 detect_stack_use_after_return=1 print_stats=1 halt_on_error=1 allocator_may_return_null=1 detect_leaks=1"
 test:asan_ubsan_lsan --test_env=UBSAN_OPTIONS="exitcode=55 allow_addr2line=1 verbosity=1 coverage=1 print_stacktrace=1 halt_on_error=1"
+# Rust: route the final link of rust_binary/rust_test through the C++ toolchain
```
arkjedrz: Those comments overexplain, just say what they're for.

```
# Enable sanitizers support for Rust.
test:asan_ubsan_lsan --@rules_rust//rust/settings:experimental_use_cc_common_link=True
test:asan_ubsan_lsan --@rules_rust//rust/settings:experimental_use_allocator_libraries_with_mangled_symbols=True
```

score/thread/thread.rs
```diff
@@ -213,20 +213,23 @@ impl<T> JoinInner<T> {
...
         // This can only be done once thread finished.
         fence(Ordering::Acquire);
-        let packet = Arc::get_mut(&mut self.packet).expect("thread not yet finished");
-        let result = packet.0.take().expect("thread result uninitialized");
-
-        // Prevent destructor from running - thread is already joined.
-        core::mem::forget(self);
-
-        result
+        let this = core::mem::ManuallyDrop::new(self);
```
arkjedrz, line 225: Restore `// Prevent destructor from running - thread is already joined.` before this line.
Remove `join_internal already joined the thread, so we must not let JoinInner::drop run it again — hence ManuallyDrop.` from the safety clause - it is misplaced and not part of unsafe operation.

score/thread/thread.rs
```diff
@@ -213,20 +213,23 @@ impl<T> JoinInner<T> {
...
+        // SAFETY: join_internal already joined the thread, so we must not let JoinInner::drop run it again
```
arkjedrz: Those generated comments are hard to track. Proposed comment:

```
SAFETY:
`self` have its `Drop` disabled, which is causing `packet` field to never clean-up.
`packet` is duplicated with `core::ptr::read` to ensure `Drop` on `Arc` is being run when going out of scope.
```

arkjedrz (changes requested):

.bazelrc
eclipse-impl: Well, that would not be completely true, see the answer to @dcalavrezo-qorix above. A complete Rust sanitizer support would require the Rust code to be instrumented, and a nightly toolchain. What we are doing is a partial enablement, of aspects achievable through linking, so just saying "Enable sanitizers support for Rust" could be misleading.

arkjedrz:
```
Partial enablement of sanitizers for Rust. Compile-time checks are not available.
Final link of `rust_binary` and `rust_test` is routed through the C++ toolchain linker.
```

Explains what is actually enabled, and what remains unavailable (regarding Dan's comment). Explains what happens underneath without going in-depth.

.bazelrc
```diff
@@ -101,14 +101,21 @@ coverage --cache_test_results=no
...
 test:asan_ubsan_lsan --features=lsan
 test:asan_ubsan_lsan --copt -g3
 test:asan_ubsan_lsan --platform_suffix=asan_ubsan_lsan
 test:asan_ubsan_lsan --test_env=ASAN_OPTIONS="exitcode=55 allow_addr2line=1 verbosity=1 coverage=1 check_initialization_order=1 detect_stack_use_after_return=1 print_stats=1 halt_on_error=1 allocator_may_return_null=1 detect_leaks=1"
 test:asan_ubsan_lsan --test_env=UBSAN_OPTIONS="exitcode=55 allow_addr2line=1 verbosity=1 coverage=1 print_stacktrace=1 halt_on_error=1"
+# Rust: route the final link of rust_binary/rust_test through the C++ toolchain
+# (cc_common.link) instead of rustc's self-contained -nodefaultlibs link. This lets
+# the --features=asan/ubsan/lsan above apply to Rust targets exactly like cc_test.
+test:asan_ubsan_lsan --@rules_rust//rust/settings:experimental_use_cc_common_link=True
+# rustc 1.90 mangles the allocator shim symbols. When cc_common.link does the final link,
```
arkjedrz: Tried to rephrase to avoid negation at the end.

```
Allow `cc_common.link` to use allocator libraries with mangled symbols, as provided by rustc 1.90.
```

arkjedrz (changes requested):

arkjedrz (approved):

PR #430, opened 2026-07-30
Add PR auto-labeler workflow

4og:
Adds a GitHub Actions workflow (`actions/labeler@v7`) that automatically applies labels to pull requests based on the files they change:

- `comp-*` component labels based on the touched `score/` subfolder
- `documentation` for changes under `docs/`
- `c++` and/or `rust` based on the type of code files changed

Mapping is defined declaratively in `.github/labeler.yml`. The workflow runs on `pull_request_target` with `sync-labels: true`.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,73 @@
+# *******************************************************************************
+# Copyright (c) 2026 Contributors to the Eclipse Foundation
+#
+# See the NOTICE file(s) distributed with this work for additional
+# information regarding copyright ownership.
+#
+# This program and the accompanying materials are made available under the
+# terms of the Apache License Version 2.0 which is available at
+# https://www.apache.org/licenses/LICENSE-2.0
+#
+# SPDX-License-Identifier: Apache-2.0
+# *******************************************************************************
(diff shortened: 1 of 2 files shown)


.github/labeler.yml
```diff
@@ -0,0 +1,73 @@
...
+# information regarding copyright ownership.
+#
+# This program and the accompanying materials are made available under the
+# terms of the Apache License Version 2.0 which is available at
+# https://www.apache.org/licenses/LICENSE-2.0
+#
+# SPDX-License-Identifier: Apache-2.0
+# *******************************************************************************
+# Component labels: applied when a PR changes files in the mapped folder.
+comp-abi-types:
```
fbaeuerle, line 14: is the comp- prefix mandated by a guideline?

fbaeuerle (approved):

.github/labeler.yml
4og: No guideline for this. It's only for baselibs to tell libraries apart in pull requests.
See https://github.com/eclipse-score/baselibs/labels (the set of labels is not complete yet)

antonkri (approved):

PR #441, opened 2026-08-03
Stop using deprecated macro in mw/log

limdor:
This is used transitively by a lot of customers and it will trigger compiler warnings to all of them without having any influence.


4og:
Thanks for PR. It just got fixed with the recent code sync: #442

PR #451, opened 2026-08-07
chore(ci): setup proper bazel caching

AlexanderLanin:


lurtz:
The workflow is not valid: https://github.com/eclipse-score/baselibs/actions/runs/31153357647

PR #452, opened 2026-08-10
os: make static_destruction_guard target publicly visible

ahmed0mousa:
Needed for other repos (e.g. communication) can reuse the nifty-counter idiom for their own static-storage singletons without vendoring a local copy.


4og:
Since it's about a general purpose utility, I propose to move it to `/score/utils`, and then make it public

4og (approved):

4og (approved):

PR #453, opened 2026-08-10
upkeep: update docs-as-code to 7.1.0

a-zw:


4og:
Hi @a-zw, thanks for the PR!
Unfortunately it doesn't work when building with bazel sandbox enabled. The git root discovery fails during the docs build and attempts to write `ubproject.toml` in the user's home folder, which is not accessible in the sandbox.

Here is the fix: https://github.com/eclipse-score/docs-as-code/pull/705

a-zw:
Set to draft until the next score_docs_as_code release is out.

4og:
@a-zw, the [v7.1.0](https://github.com/eclipse-score/docs-as-code/releases/tag/v7.1.0) has just been released.

4og (approved):

PR #461, opened 2026-08-11
Add Rust wrapper for score::cpp::stop_token

eclipse-impl:
Add Rust wrapper around `score::cpp::stop_token`, allowing Rust applications to interact with frameworks that require stop tokens, such as lifecycle.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,6 +26,7 @@ default-members = [
     "score/testing_macros",
     "score/language/rust/libcpp",
     "score/language/rust/libcpp_derive",
+    "score/language/rust/stop_token",
     "score/result/rust",
     "score/language/rust/memres",
 ]
@@ -44,6 +45,7 @@ members = [
     "score/testing_macros",
     "score/language/rust/libcpp",
     "score/language/rust/libcpp_derive",
+    "score/language/rust/stop_token",
     "score/result/rust",
     "examples/log_builtin",
     "examples/log_custom",
--- a/score/language/rust/stop_token/BUILD
+++ b/score/language/rust/stop_token/BUILD
@@ -1,0 +1,54 @@
+# *******************************************************************************
+# Copyright (c) 2026 Contributors to the Eclipse Foundation
+#
+# See the NOTICE file(s) distributed with this work for additional
+# information regarding copyright ownership.
+#
+# This program and the accompanying materials are made available under the
+# terms of the Apache License Version 2.0 which is available at
+# https://www.apache.org/licenses/LICENSE-2.0
+#
+# SPDX-License-Identifier: Apache-2.0
+# *******************************************************************************
+
+load("@cxx.rs//tools/bazel:rust_cxx_bridge.bzl", "rust_cxx_bridge")
+load("@rules_cc//cc:defs.bzl", "cc_library")
+load("@rules_rust//rust:defs.bzl", "rust_library", "rust_test")
+load("@score_baselibs//score/language/safecpp:toolchain_features.bzl", "COMPILER_WARNING_FEATURES")
+
+cc_library(
+    name = "stop_token_adapter",
+    srcs = ["cpp/stop_token_adapter.cpp"],
+    hdrs = ["cpp/stop_token_adapter.h"],
+    features = COMPILER_WARNING_FEATURES,
+    deps = ["@score_baselibs//score/language/futurecpp"],
+)
+
+rust_cxx_bridge(
+    name = "stop_token_bridge",
+    src = "src/lib.rs",
+    deps = [":stop_token_adapter"],
+)
+
+rust_library(
+    name = "stop_token",
+    srcs = ["src/lib.rs"],
+    visibility = ["//visibility:public"],
+    deps = [
+        ":stop_token_bridge",
+        "@cxx.rs//:cxx",
+    ],
+)
+
+alias(
+    name = "rust",
+    actual = ":stop_token",
+    visibility = ["//visibility:public"],
+)
+
+rust_test(
+    name = "stop_token_test",
+    crate = ":stop_token",
+    features = ["link_std_cpp_lib"],
+    tags = ["unit"],
+)
--- a/score/language/rust/stop_token/BUILD
+++ b/score/language/rust/stop_token/BUILD
@@ -40,12 +40,6 @@ rust_library(
     ],
 )
 
-alias(
-    name = "rust",
-    actual = ":stop_token",
-    visibility = ["//visibility:public"],
-)
-
 rust_test(
     name = "stop_token_test",
     crate = ":stop_token",
--- a/score/language/rust/stop_token/Cargo.toml
+++ b/score/language/rust/stop_token/Cargo.toml
@@ -1,0 +1,30 @@
+# *******************************************************************************
+# Copyright (c) 2026 Contributors to the Eclipse Foundation
+#
+# See the NOTICE file(s) distributed with this work for additional
+# information regarding copyright ownership.
+#
+# This program and the accompanying materials are made available under the
+# terms of the Apache License Version 2.0 which is available at
+# https://www.apache.org/licenses/LICENSE-2.0
+#
+# SPDX-License-Identifier: Apache-2.0
+# *******************************************************************************
+
+[package]
+name = "stop_token"
+description = "Rust bindings for score::cpp::stop_token and score::cpp::stop_source."
+version.workspace = true
+authors.workspace = true
+readme.workspace = true
+edition.workspace = true
+license-file.workspace = true
+
+[lib]
+path = "src/lib.rs"
+
+[dependencies]
+cxx = "1.0.186"
+
+[lints]
+workspace = true
--- a/score/language/rust/stop_token/Cargo.toml
+++ b/score/language/rust/stop_token/Cargo.toml
@@ -26,5 +26,8 @@ path = "src/lib.rs"
 [dependencies]
 cxx = "1.0.186"
 
+[build-dependencies]
+cxx-build = "1.0.186"
+
 [lints]
 workspace = true
--- a/score/language/rust/stop_token/cpp/stop_token_adapter.cpp
+++ b/score/language/rust/stop_token/cpp/stop_token_adapter.cpp
@@ -47,9 +47,9 @@ bool StopSourceRequestStop(const score::cpp::stop_source& source) noexcept
     return const_cast<score::cpp::stop_source&>(source).request_stop();
 }
 
-std::shared_ptr<score::cpp::stop_token> StopSourceGetToken(const score::cpp::stop_source& source) noexcept
+std::unique_ptr<score::cpp::stop_token> StopSourceGetToken(const score::cpp::stop_source& source) noexcept
 {
-    return std::make_shared<score::cpp::stop_token>(source.get_token());
+    return std::make_unique<score::cpp::stop_token>(source.get_token());
 }
 
 bool StopSourceEqual(const score::cpp::stop_source& lhs, const score::cpp::stop_source& rhs) noexcept
@@ -57,14 +57,14 @@ bool StopSourceEqual(const score::cpp::stop_source& lhs, const score::cpp::stop_
     return lhs == rhs;
 }
 
-std::shared_ptr<score::cpp::stop_token> MakeDefaultStopToken() noexcept
+std::unique_ptr<score::cpp::stop_token> MakeStopToken() noexcept
 {
-    return std::make_shared<score::cpp::stop_token>();
+    return std::make_unique<score::cpp::stop_token>();
 }
 
-std::shared_ptr<score::cpp::stop_source> MakeStopSource() noexcept
+std::unique_ptr<score::cpp::stop_source> MakeStopSource() noexcept
 {
-    return std::make_shared<score::cpp::stop_source>();
+    return std::make_unique<score::cpp::stop_source>();
 }
 
 }  // namespace score::language::rust::stop_token
--- a/score/language/rust/stop_token/cpp/stop_token_adapter.cpp
+++ b/score/language/rust/stop_token/cpp/stop_token_adapter.cpp
@@ -1,0 +1,70 @@
+/*******************************************************************************
+ * Copyright (c) 2026 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Apache License Version 2.0 which is available at
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ *******************************************************************************/
+
+#include "score/language/rust/stop_token/cpp/stop_token_adapter.h"
+
+namespace score::language::rust::stop_token
+{
+
+bool StopTokenStopRequested(const score::cpp::stop_token& token) noexcept
+{
+    return token.stop_requested();
+}
+
+bool StopTokenStopPossible(const score::cpp::stop_token& token) noexcept
+{
+    return token.stop_possible();
+}
+
+bool StopTokenEqual(const score::cpp::stop_token& lhs, const score::cpp::stop_token& rhs) noexcept
+{
+    return lhs == rhs;
+}
+
+bool StopSourceStopRequested(const score::cpp::stop_source& source) noexcept
+{
+    return source.stop_requested();
+}
+
+bool StopSourceStopPossible(const score::cpp::stop_source& source) noexcept
+{
+    return source.stop_possible();
+}
+
+bool StopSourceRequestStop(const score::cpp::stop_source& source) noexcept
+{
+    // request_stop() only updates the shared stop state atomically.
+    return const_cast<score::cpp::stop_source&>(source).request_stop();
+}
+
+std::shared_ptr<score::cpp::stop_token> StopSourceGetToken(const score::cpp::stop_source& source) noexcept
+{
+    return std::make_shared<score::cpp::stop_token>(source.get_token());
+}
+
+bool StopSourceEqual(const score::cpp::stop_source& lhs, const score::cpp::stop_source& rhs) noexcept
+{
+    return lhs == rhs;
+}
+
+std::shared_ptr<score::cpp::stop_token> MakeDefaultStopToken() noexcept
+{
+    return std::make_shared<score::cpp::stop_token>();
+}
+
+std::shared_ptr<score::cpp::stop_source> MakeStopSource() noexcept
+{
+    return std::make_shared<score::cpp::stop_source>();
+}
+
+}  // namespace score::language::rust::stop_token
--- a/score/language/rust/stop_token/cpp/stop_token_adapter.h
+++ b/score/language/rust/stop_token/cpp/stop_token_adapter.h
@@ -40,16 +40,16 @@ bool StopSourceStopPossible(const score::cpp::stop_source& source) noexcept;
 bool StopSourceRequestStop(const score::cpp::stop_source& source) noexcept;
 
 /// @brief Returns a token associated with a stop source.
-std::shared_ptr<score::cpp::stop_token> StopSourceGetToken(const score::cpp::stop_source& source) noexcept;
+std::unique_ptr<score::cpp::stop_token> StopSourceGetToken(const score::cpp::stop_source& source) noexcept;
 
 /// @brief Returns whether two stop sources share the same stop state.
 bool StopSourceEqual(const score::cpp::stop_source& lhs, const score::cpp::stop_source& rhs) noexcept;
 
 /// @brief Creates a default-constructed stop token.
-std::shared_ptr<score::cpp::stop_token> MakeDefaultStopToken() noexcept;
+std::unique_ptr<score::cpp::stop_token> MakeStopToken() noexcept;
 
 /// @brief Creates a stop source with a new stop state.
-std::shared_ptr<score::cpp::stop_source> MakeStopSource() noexcept;
+std::unique_ptr<score::cpp::stop_source> MakeStopSource() noexcept;
 
 }  // namespace score::language::rust::stop_token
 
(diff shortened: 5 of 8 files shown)


Cargo.toml
```diff
@@ -26,6 +26,7 @@ default-members = [
     "score/testing_macros",
     "score/language/rust/libcpp",
     "score/language/rust/libcpp_derive",
+    "score/language/rust/stop_token",
```
arkjedrz, line 29: What happens when You `cargo build` the project?

score/language/rust/stop_token/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+authors.workspace = true
+readme.workspace = true
+edition.workspace = true
+license-file.workspace = true
+
+[lib]
+path = "src/lib.rs"
+
+[dependencies]
+cxx = "1.0.186"
```
arkjedrz, line 27: Replace with a regular FFI layer.

arkjedrz (changes requested):

Cargo.toml
eclipse-impl: it works:
```
$ cargo build
   Compiling score_log_fmt v0.0.1 (/home/q474126/code/eclipse-score-baselibs/score/log_rust/score_log_fmt)
   Compiling foldhash v0.2.0
   Compiling memres v0.0.1 (/home/q474126/code/eclipse-score-baselibs/score/language/rust/memres)
   Compiling libc v0.2.189
   Compiling syn v2.0.119
   Compiling link-cplusplus v1.0.12
warning: missing documentation for a struct field
   --> score/language/rust/memres/src/lib.rs:141:17
    |
141 |     Monotonic { size: usize },
    |                 ^^^^^^^^^^^
    |
    = note: requested on the command line with `-W missing-docs`

warning: `memres` (lib) generated 1 warning
   Compiling cxx v1.0.199
   Compiling stop_token v0.0.1 (/home/q474126/code/eclipse-score-baselibs/score/language/rust/stop_token)
   Compiling score_log_fmt_macro v0.0.1 (/home/q474126/code/eclipse-score-baselibs/score/log_rust/score_log_fmt_macro)
   Compiling libcpp_derive v0.0.1 (/home/q474126/code/eclipse-score-baselibs/score/language/rust/libcpp_derive)
warning: missing documentation for the crate
   --> score/language/rust/libcpp_derive/lib.rs:14:1
    |
 14 | / use proc_macro2::TokenStream;
 15 | | use quote::{format_ident, quote};
 16 | | use syn::parse::Parse;
 17 | | use syn::spanned::Spanned;
...   |
234 | | }
    | |_^
    |
    = note: requested on the command line with `-W missing-docs`

warning: `libcpp_derive` (lib) generated 1 warning
   Compiling libcpp v0.0.1 (/home/q474126/code/eclipse-score-baselibs/score/language/rust/libcpp)
   Compiling score_log v0.0.1 (/home/q474126/code/eclipse-score-baselibs/score/log_rust/score_log)
   Compiling result_rs v0.0.1 (/home/q474126/code/eclipse-score-baselibs/score/result/rust)
warning: missing documentation for a struct
   --> score/result/rust/result.rs:156:5
    |
156 |     pub struct Expected<T, E> {
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: requested on the command line with `-W missing-docs`

   Compiling allocator v0.0.1 (/home/q474126/code/eclipse-score-baselibs/score/allocator)
   Compiling stdout_logger v0.0.1 (/home/q474126/code/eclipse-score-baselibs/score/log_rust/stdout_logger)
   Compiling pal v0.0.1 (/home/q474126/code/eclipse-score-baselibs/score/pal)
warning: `result_rs` (lib) generated 1 warning
   Compiling sync v0.0.1 (/home/q474126/code/eclipse-score-baselibs/score/sync)
   Compiling containers v0.0.1 (/home/q474126/code/eclipse-score-baselibs/score/containers_rust)
   Compiling thread v0.0.1 (/home/q474126/code/eclipse-score-baselibs/score/thread)
   Compiling score_testing_macros v0.0.1 (/home/q474126/code/eclipse-score-baselibs/score/testing_macros)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 2.65s
```

score/language/rust/stop_token/src/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+// information regarding copyright ownership.
+//
+// This program and the accompanying materials are made available under the
+// terms of the Apache License Version 2.0 which is available at
+// https://www.apache.org/licenses/LICENSE-2.0
+//
+// SPDX-License-Identifier: Apache-2.0
+// *******************************************************************************
+
+//! Lifetime-safe Rust bindings for C++ `score::cpp::stop_token` and
```
arkjedrz: Unnecessary wrap.

```rust
//! Lifetime-safe Rust bindings for C++ `score::cpp::stop_token` and `score::cpp::stop_source`.
```

score/language/rust/stop_token/BUILD
```diff
@@ -0,0 +1,54 @@
...
+    name = "stop_token",
+    srcs = ["src/lib.rs"],
+    visibility = ["//visibility:public"],
+    deps = [
+        ":stop_token_bridge",
+        "@cxx.rs//:cxx",
+    ],
+)
+
+alias(
```
arkjedrz: Unnecessary alias? Why need "//score/language/<ins>rust</ins>/stop_token:<ins>rust</ins>", when "//score/language/rust/stop_token" can be used?

score/language/rust/stop_token/BUILD
```diff
@@ -0,0 +1,54 @@
...
+
+cc_library(
+    name = "stop_token_adapter",
+    srcs = ["cpp/stop_token_adapter.cpp"],
+    hdrs = ["cpp/stop_token_adapter.h"],
+    features = COMPILER_WARNING_FEATURES,
+    deps = ["@score_baselibs//score/language/futurecpp"],
+)
+
+rust_cxx_bridge(
```
arkjedrz, line 27: I might be missing an understanding regarding how `cxx` works - correct me if I'm wrong.
Isn't `src/lib.rs` used twice? Shouldn't it be like this?
- Move `mod ffi` content to a separate `ffi.rs`/`cxx.rs` file.
- Link this file as `src` to the bridge (here).
- Link `stop_token_bridge` to `stop_token` as a dependency.

score/language/rust/stop_token/src/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+
+/// Creates an owned source with a new associated stop state.
+pub fn make_stop_source() -> cxx::SharedPtr<StopSource> {
+    ffi::make_stop_source()
+}
+
+unsafe impl Send for StopToken {}
+unsafe impl Sync for StopToken {}
+unsafe impl Send for StopSource {}
+unsafe impl Sync for StopSource {}
```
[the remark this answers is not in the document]
arkjedrz: Agreed, add safety clauses everywhere `unsafe` is used.

score/language/rust/stop_token/src/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+        fn make_stop_source() -> SharedPtr<stop_source>;
+    }
+
+    // This forces CXX to generate glue code for the shared pointers.
+    impl SharedPtr<stop_token> {}
+    impl SharedPtr<stop_source> {}
+}
+
+/// Opaque view of C++ `score::cpp::stop_token`.
+pub use ffi::stop_token as StopToken;
```
arkjedrz: This is not opaque, make it a tuple struct. Right now it's an alias.

```rust
pub struct StopToken(ffi::stop_token);
```

score/language/rust/stop_token/src/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+
+/// Opaque view of C++ `score::cpp::stop_source`.
+pub use ffi::stop_source as StopSource;
+
+/// Creates an owned, default-constructed token without associated stop state.
+pub fn make_default_stop_token() -> cxx::SharedPtr<StopToken> {
```
arkjedrz: Convert to a factory method of `StopToken`.

score/language/rust/stop_token/src/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+    // This forces CXX to generate glue code for the shared pointers.
+    impl SharedPtr<stop_token> {}
+    impl SharedPtr<stop_source> {}
+}
+
+/// Opaque view of C++ `score::cpp::stop_token`.
+pub use ffi::stop_token as StopToken;
+
+/// Opaque view of C++ `score::cpp::stop_source`.
+pub use ffi::stop_source as StopSource;
```
arkjedrz: Make it a tuple struct.

score/language/rust/stop_token/cpp/stop_token_adapter.cpp
```diff
@@ -0,0 +1,70 @@
...
+{
+    return std::make_shared<score::cpp::stop_token>(source.get_token());
+}
+
+bool StopSourceEqual(const score::cpp::stop_source& lhs, const score::cpp::stop_source& rhs) noexcept
+{
+    return lhs == rhs;
+}
+
+std::shared_ptr<score::cpp::stop_token> MakeDefaultStopToken() noexcept
```
arkjedrz: `MakeDefaultStopToken` -> `MakeStopToken`? What's particularly default about it?

score/language/rust/stop_token/src/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+
+/// Creates an owned, default-constructed token without associated stop state.
+pub fn make_default_stop_token() -> cxx::SharedPtr<StopToken> {
+    ffi::make_default_stop_token()
+}
+
+/// Creates an owned source with a new associated stop state.
+pub fn make_stop_source() -> cxx::SharedPtr<StopSource> {
```
arkjedrz: Convert to a factory method of `StopSource`.

score/language/rust/stop_token/src/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+    }
+}
+
+impl PartialEq for StopToken {
+    fn eq(&self, other: &Self) -> bool {
+        ffi::stop_token_equal(self, other)
+    }
+}
+
+impl Eq for StopToken {}
```
arkjedrz: Use derive macro?

score/language/rust/stop_token/src/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+        #[cxx_name = "MakeDefaultStopToken"]
+        fn make_default_stop_token() -> SharedPtr<stop_token>;
+
+        #[namespace = "score::language::rust::stop_token"]
+        #[cxx_name = "MakeStopSource"]
+        fn make_stop_source() -> SharedPtr<stop_source>;
+    }
+
+    // This forces CXX to generate glue code for the shared pointers.
+    impl SharedPtr<stop_token> {}
```
arkjedrz: Why `shared_ptr`, and not just heap allocation and destroy on `Drop`? Those internal objects have a single owner.

arkjedrz (changes requested):

score/language/rust/stop_token/BUILD
eclipse-impl: Yes, the file is processed twice, one for the contents of `#[cxx::bridge]`, that will generate the C++ glue code, and once for the remainder of the file. One does not need to separate the usages on different files, and both the [tutorial](https://cxx.rs/tutorial.html) and our other modules do it on the same file.

score/language/rust/stop_token/src/lib.rs
eclipse-impl: I don't totally follow your comment, yes, it's an alias, but from Rust's PoV it's opaque because Rust does not know the structure of the object behind it, and can only use references to it through the provided C++ API.
Having the tuple struct would mean that we have to create code to pass the `.0` element everywhere. What would be the benefit that you see in that?

score/language/rust/stop_token/BUILD
eclipse-impl: Indeed. Removed.

score/language/rust/stop_token/cpp/stop_token_adapter.cpp
eclipse-impl: Indeed, on the open source version, this makes no sense.

score/language/rust/stop_token/src/lib.rs
eclipse-impl: A derive macro is not possible as we only have the alias, but even if we'd wrap it into a struct, the automatic derivation would not work, as the underlying FFI type does not support it, so for interacting with FFI code, so we'd have to provide it anyway.

score/language/rust/stop_token/src/lib.rs
eclipse-impl: Valid point. I didn't find any API that explicitly requires the `shared_ptr`, a `unique_ptr` would do. We still need it, though, because the heap allocation happens at C++ side, so C++ has to clean-it-up. Sure, we could have a "fat" wrapper just to make sure we call C++ to destroy the object on drop, but we can get that behaviour for free with the `unique_ptr`. Given the explanation above that using the alias avoids a lot of boilerplate code just to deal with `.0` in a struct, I'd argue to change the API to `unique_ptr` but not to manually do this management.

score/language/rust/stop_token/src/lib.rs
eclipse-impl: As explained above, this would require extra glue code without any obvious benefit. Maybe you can elaborate on the reasons why that would be necessary?

score/language/rust/stop_token/src/lib.rs
eclipse-impl: Fixed.

score/language/rust/stop_token/src/lib.rs
arkjedrz: From library consumer PoV You're leaking implementation details of `UniquePtr`, which should just be hidden away from the consumer.

Take a look at unit tests - the usage is getting quirky very quickly. You're enforcing every consumer to obtain a reference (`Option<&Type>`) then check if it actually allocated. More expected behavior would be to have a constructor that panics if for some reason pointer is null. This is how it'd look like in a test.

```rust
#[test]
fn default_token_has_no_stop_state() {
    let token = StopToken::new();

    assert!(!token.stop_requested());
    assert!(!token.stop_possible());
}
```

Newtype is a common Rust pattern for such cases. In this case it's not a matter of being idiomatic, but a matter of API correctness.

score/language/rust/stop_token/src/lib.rs
```diff
@@ -0,0 +1,178 @@
...
+/// SAFETY: StopSource is a means to share a `stop_state` and that class uses atomics to guarantee
+/// that the state is consistent between threads, as described at:
+/// score/language/futurecpp/include/score/private/stop_token/stop_state.hpp
+/// Therefore it is safe to move and share these objects between threads.
+unsafe impl Send for StopSource {}
+unsafe impl Sync for StopSource {}
+
+impl StopToken {
+    /// Creates an owned, default-constructed token without associated stop state.
+    pub fn make() -> cxx::UniquePtr<Self> {
```
arkjedrz: 1) `new` would be the idiomatic name for factory method.
2) With newtype/tuple struct this method would return more obvious `Self` type.

score/language/rust/stop_token/src/lib.rs
```diff
@@ -0,0 +1,178 @@
...
+        ffi::stop_source_stop_possible(self)
+    }
+
+    /// Requests stop on the associated stop state.
+    pub fn request_stop(&self) -> bool {
+        ffi::stop_source_request_stop(self)
+    }
+
+    /// Returns an owned token associated with this source's stop state.
+    pub fn get_token(&self) -> cxx::UniquePtr<StopToken> {
```
arkjedrz: With newtype/tuple struct this method would return more obvious `StopToken` type.

score/language/rust/stop_token/cpp/stop_token_adapter.h
```diff
@@ -0,0 +1,56 @@
...
+ * terms of the Apache License Version 2.0 which is available at
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ *******************************************************************************/
+
+#ifndef SCORE_LANGUAGE_RUST_STOP_TOKEN_CPP_STOP_TOKEN_ADAPTER_H
+#define SCORE_LANGUAGE_RUST_STOP_TOKEN_CPP_STOP_TOKEN_ADAPTER_H
+
+#include <score/stop_token.hpp>
```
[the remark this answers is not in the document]
arkjedrz: Follow the guide.

score/language/rust/stop_token/cpp/stop_token_adapter.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+bool StopSourceStopPossible(const score::cpp::stop_source& source) noexcept
+{
+    return source.stop_possible();
+}
+
+bool StopSourceRequestStop(const score::cpp::stop_source& source) noexcept
+{
+    // request_stop() only updates the shared stop state atomically.
+    return const_cast<score::cpp::stop_source&>(source).request_stop();
```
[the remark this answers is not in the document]
arkjedrz: Can `StopSourceRequestStop` accept `score::cpp::stop_source& source` instead of `const ...`?

score/language/rust/stop_token/src/lib.rs
```diff
@@ -0,0 +1,178 @@
...
+impl Eq for StopSource {}
+
+#[cfg(test)]
+mod tests {
+    use super::{StopSource, StopToken};
+
+    #[test]
+    fn default_token_has_no_stop_state() {
+        let token = StopToken::make();
+        let token = token.as_ref().expect("default token must be allocated");
```
arkjedrz: As mentioned in other comment - both those lines should just be:

```rust
let token = StopToken::new();
```

score/language/rust/stop_token/src/lib.rs
```diff
@@ -0,0 +1,178 @@
...
+
+        assert!(!token.stop_requested());
+        assert!(!token.stop_possible());
+    }
+
+    #[test]
+    fn source_requests_stop_visible_to_its_token() {
+        let source = StopSource::make();
```
arkjedrz: Those lines should be replaced with:

```rust
let source = StopSource::new();
let token = source.get_token();
```

arkjedrz (changes requested):

score/language/rust/stop_token/cpp/stop_token_adapter.cpp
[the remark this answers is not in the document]
eclipse-impl: That would require:
```rust
    pub fn stop_requested(&mut self) -> bool {
        ffi::stop_token_stop_requested(self)
    }
```
And that would be terrible, as only one thread can hold a mutable reference to an object at a time. This would make multi-threading way more complicated than what it should. As this object is meant to be used on MT applications, we discussed internally that removing the `const` in the C++ side is the correct approach, as this is safe to do (only atomics are involved) and they allow for the proper semantics on the Rust side.

score/language/rust/stop_token/src/lib.rs
eclipse-impl: I see your point. Indeed. I'm a bit afraid that this will make the reintegration on our codebase a bit more complex, but you're right that the API is not correct. I'll update this.

score/language/rust/stop_token/src/lib.rs
```diff
@@ -0,0 +1,212 @@
...
+
+impl StopToken {
+    /// Creates an owned, default-constructed token without associated stop state.
+    pub fn new() -> Self {
+        Self::from_inner(ffi::make_stop_token())
+    }
+
+    fn from_inner(inner: cxx::UniquePtr<ffi::stop_token>) -> Self {
+        if inner.is_null() {
+            std::process::abort();
```
arkjedrz, line 100: Use `panic!` with a reason message.

score/language/rust/stop_token/src/lib.rs
```diff
@@ -0,0 +1,212 @@
...
+        }
+
+        Self(inner)
+    }
+
+    fn as_ffi(&self) -> &ffi::stop_token {
```
arkjedrz, line 106: This method is not necessary. Other methods can just use `&self.0`.

This shouldn't matter after removal, but `unwrap_or_else(|| std::process::abort())` is just an `expect()` without reason message.

score/language/rust/stop_token/src/lib.rs
```diff
@@ -0,0 +1,212 @@
...
+        ffi::stop_token_stop_requested(self.as_ffi())
+    }
+
+    /// Returns whether this token has associated stop state that can be stopped.
+    pub fn stop_possible(&self) -> bool {
+        ffi::stop_token_stop_possible(self.as_ffi())
+    }
+}
+
+impl Default for StopToken {
```
arkjedrz, line 121: Isn't this exactly how `Default` derive macro would look like? I might be mistaken.

score/language/rust/stop_token/src/lib.rs
```diff
@@ -0,0 +1,212 @@
...
+    }
+
+    /// Requests stop on the associated stop state.
+    pub fn request_stop(&self) -> bool {
+        ffi::stop_source_request_stop(self.as_ffi())
+    }
+
+    /// Returns an owned token associated with this source's stop state.
+    pub fn get_token(&self) -> StopToken {
+        StopToken::from_inner(ffi::stop_source_get_token(self.as_ffi()))
```
arkjedrz, line 179: Maybe even `from_inner` is not required. Following seems to work. This would make `from_inner` factory methods collapse into `new` methods.

```rust
StopToken(ffi::stop_source_get_token(&self.0))
```

arkjedrz (changes requested):

score/language/rust/stop_token/src/lib.rs
eclipse-impl: The reason for the method was to exactly handle the case when the value cannot be unwrapped the same way everywhere (guarantee the consistency).

score/language/rust/stop_token/src/lib.rs
eclipse-impl: This would not be the same. The macro would expand to:
```rust
impl Default for StopSource {
    fn default() -> Self {
        StopSource(cxx::UniquePtr::<ffi::stop_source>::default())
    }
}
```
For a `UniquePtr` the default is to return a nullptr, which is not what we want here.

Cargo.toml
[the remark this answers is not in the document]
eclipse-impl: Added `build.rs` file.

arkjedrz (dismissed):
LGTM, but fix CI issues.

arkjedrz (approved):

PR #464, opened 2026-08-12
docs: move component docs from docs/baselibs/components/ to score/*/docs

### LINKED ISSUE #235: Move documentation closer to code
Currently, the module documentation is not located alongside the relevant code as recommended. To improve maintainability and discoverability, documentation should be relocated so it's placed directly with the corresponding components in the codebase (i.e. `/score/<component>/docs`).

File layout is described in the module template: https://eclipse-score.github.io/module_template/main/template/index.html

Rutik7:
Closes #235


BUILD
```diff
@@ -23,7 +23,7 @@ docs(
         "@score_platform//:needs_json",
         "@score_process//:needs_json",
     ],
-    source_dir = "docs",
+    source_dir = ".",
```
4og: I think this should remain "docs".

`score_lifecycle` is a good example how this to be done. See https://github.com/eclipse-score/lifecycle/blob/main/BUILD

score/concurrency/BUILD
```diff
@@ -632,3 +633,9 @@ cc_unit_test_suites_for_host_and_qnx(
         "//visibility:public",
     ],
 )
+
+docs_bundle(
```
4og, line 637: The integration tests are failing on CI because the `docs_bundle` is located in `score_docs_as_code`, which is a dev-dependency. To fix it, I think each `docs/` folder should have a BUILD file with the `docs_bundle` declaration. @AlexanderLanin, would this work?

4og (changes requested):

AlexanderLanin:
To my knowledge all of these must be non dev dependencies. That's how all other modules solve it.

```
bazel_dep(name = "score_docs_as_code", version = "7.1.0", dev_dependency = True)
bazel_dep(name = "score_process", version = "2.0.3", dev_dependency = True)
bazel_dep(name = "score_platform", version = "0.6.4", dev_dependency = True)
```

4og:
Looks good to me. Can you address the conflicts please?

AlexanderLanin:
I'd recommend you move all the BUILD files with docs_bundle one level higher outside the respective docs dirs. I'm hoping that's a 2 minute job here at the moment, but this will prevent future large scale changes.

Since we expect docs_bundle will soon accept additional parameters like code_targets (your cc_library under documentation) and tests (your cc_unit_tests for the respective component).

4og (approved):

PR #467, opened 2026-08-12
feat(json): propagate field name in FromJsonAny deserialization errors

AAmbuj:
JsonDeserializeStructImpl now annotates every deserialization failure with the name of the struct field that caused it, stored in the error's UserMessage(). Previously, field identity was lost and only a generic error code was returned.

- Type mismatch: UserMessage() = failing field name (e.g. "pdu_id")
- Missing mandatory field: UserMessage() = absent field name
- Nested struct failure: UserMessage() = outermost containing field name

field_name is a static const char* from struct_visitable, so no heap allocation occurs and no string_view lifetime issue is introduced.

Add three tests covering type mismatch, missing key, and nested struct error propagation cases.

### DIFF
--- a/score/json/json_serializer.h
+++ b/score/json/json_serializer.h
@@ -205,7 +205,8 @@ inline void JsonDeserializeStructImpl(DeserializeAsJson& visitor, Field& field, 
         }
         else
         {
-            score::cpp::ignore = visitor.error.emplace(std::move(field_content).error());
+            // Annotate with the failing field name; field_name is a static const char* from struct_visitable.
+            score::cpp::ignore = visitor.error.emplace(field_content.error().WithUserMessage(field_name));
         }
     }
     else
@@ -220,7 +221,7 @@ inline void JsonDeserializeStructImpl(DeserializeAsJson& visitor, Field& field, 
         }
         else
         {
-            score::cpp::ignore = visitor.error.emplace(Error::kKeyNotFound, "Missing mandatory field in JSON object");
+            score::cpp::ignore = visitor.error.emplace(MakeError(Error::kKeyNotFound, field_name));
         }
     }
 }
--- a/score/json/json_serializer_test.cpp
+++ b/score/json/json_serializer_test.cpp
@@ -719,5 +719,90 @@ TEST(JsonSerializerTest, UseCustomSerializationOnVisitableStruct)
     EXPECT_EQ(serialized.As<Object>()->get().count("foo"), 1);
 }
 
+TEST(JsonSerializerTest, TypeMismatchErrorContainsFieldName)
+{
+    RecordProperty("TestType", "interface-test");
+    RecordProperty("Verifies", "::score::json::FromJsonAny");
+    RecordProperty("Description",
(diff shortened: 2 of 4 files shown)


score/json/json_serializer.h
```diff
@@ -205,7 +205,9 @@ inline void JsonDeserializeStructImpl(DeserializeAsJson& visitor, Field& field,
         }
         else
         {
-            score::cpp::ignore = visitor.error.emplace(std::move(field_content).error());
+            // Annotate with the failing field name; field_name is a static const char* from struct_visitable.
+            score::cpp::ignore = visitor.error.emplace(
+                MakeError(static_cast<Error>(*field_content.error()), field_name));
```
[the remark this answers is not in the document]
AAmbuj: Fixed. Added Error::WithUserMessage() to error.h which returns a copy preserving the original numeric code and ErrorDomain*, replacing only user_messages_. JsonDeserializeStructImpl now calls field_content.error().WithUserMessage(field_name) instead of MakeError(static_cast<json::Error>(...))json::Error(...)), so a TypedHash field failure carrying score::hash::ErrorCode::kInvalidParameters propagates with its hash domain intact.

score/json/json_serializer_test.cpp
```diff
@@ -719,5 +719,90 @@ TEST(JsonSerializerTest, UseCustomSerializationOnVisitableStruct)
...
+    "nested_int": 43,
+    "nested_bool": "not-a-bool",
+    "nested_array": [44, 45]
+    }
+} )"_json;
+
+    // When deserializing into the struct
+    auto unit{FromJsonAny<TypeToSerialize>(std::move(source))};
+
+    // Then the error code is WrongType and the UserMessage names the parent struct field, not the inner field
```
Nikhil2206, line 801: Shouldn't we add error for specific error field in nested struct, ie. "nested_bool" , "next_int" etc. 
instead of "nested_value" which is again kind of generic error ?

AAmbuj: Hi @Nikhil2206 thanks for the review, for my side I have added some points.

The serializer re-stamps the error with the current field name at each struct level (see json_serializer.h line 209: field_content.error().WithUserMessage(field_name)), so callers always get the top-level failing field of the struct they directly deserialized.

Two reasons we keep it this way:

- Stable error contract - if callers inspect UserMessage() for error handling or logging, exposing leaf names means renaming a field inside a nested type becomes a silent breaking change for all error consumers.

- Encapsulation - reporting nested_val.nested_bool leaks the internal shape of NestedType as externally observable behavior.

let me know if you want to discuss more?

AAmbuj:
Hi @fbaeuerle and @4og  could you please review this PR when you have a moment? Your feedback would be greatly appreciated.

score/json/json_serializer_test.cpp
Nikhil2206: Alright for now, however we can check the consumer of these logs and make sure it's not breaking then we can have opportunity to get that more detailed error message (the original intent of this work).

4og (approved):

PR #471, opened 2026-08-13
Remove //score/datetime_converter target and functionality

limdor:
The implementation is error prone, it had already several bugs in the past and at least I could find another one that basically it does not work proper with leap seconds.
This is not yet widely used, before we get more people depending on it and get unexpected bugs, we delete it.
If the functionality would be needed in the future, it would be better to rely as much as possible in existing implementations like some of the POSIX / C lib apis.


4og:
Hi @Rutuja-Patil-Bosch, @srinivasugithub, @gierer,

You have contributed to this code recently, so we thought we'd check with you before removing it. Are you aware of any remaining use cases, or do you have any concerns about deleting it?

gierer:
> Hi @Rutuja-Patil-Bosch, @srinivasugithub, @gierer,
> 
> You have contributed to this code recently, so we thought we'd check with you before removing it. Are you aware of any remaining use cases, or do you have any concerns about deleting it?

I'm fine with removing it.

4og (approved):

PR #491, opened 2026-08-19
demo: render module verification report via score_module_verification_report (tracks docs-as-code#730)

antonkri:


docs/reporting/module_verification_report.rst
pahmann, line 1: The document which is the rendered HTML of this rst file looks different to the structure of the module verification report template. This is already existing. 

In other words, this document https://eclipse-score.github.io/baselibs/pr-491/reporting/module_verification_report.html does not match the structure existing in the module and process definition like here: https://eclipse-score.github.io/baselibs/pr-491/verification_report/module_verification_report.html

This relate to the following file as well: https://github.com/eclipse-score/docs-as-code/pull/730/changes#diff-476160124f11cd0e0a38c257375ef66bd5d406a587ce15f92986546948400e32

pahmann (changes requested):

BUILD
```diff
@@ -96,6 +97,28 @@ docs(
...
+# validate the docs_and_test integration and coverage report rendering.
+#
+# Extra Bazel flags (e.g. --config=…) go on the command line after --, e.g.::
+#
+#     bazel run //:module_verification_report -- --test-flag=--config=bl-aarch64-linux
+#
+# See @score_docs_as_code//:bzl/docs_and_test.bzl for the underlying driver.
+docs_and_test(
+    name = "module_verification_report",
+    test_targets = [
```
PiotrKorkus, line 114: I imagine different test targets might require different configurations e.g. ITF vs UT. How this can be handled when we have a single target to execute everything in one place?

Can we leave `test_targets` empty if we want to handle execution of tests ourselves and just use it for rendering module verification?

PR #499, opened 2026-08-21
bitmanipulation: link unit tests to comp_req requirements

4og:
Link all `bit_manipulation` and `bitmask_operators` unit tests to the `comp_req__bitmanipulation__*` requirements they verify, using `FullyVerifies`/`PartiallyVerifies`, `TestType`, `DerivationTechnique`, and `Description` metadata per the `test-to-requirement-linking` skill.

Also sharpens the skill's own guidance:
- Tightens the `equivalence-classes` vs `boundary-values` distinction with a concrete decision rule.
- Adds a `TestType` breakdown table (previously only `DerivationTechnique` had one).
- Notes that a passing `bazel test` does not confirm a linked requirement ID actually resolves; only `bazel run //:docs` validates that.

### DIFF
--- a/score/bitmanipulation/bit_manipulation_test.cpp
+++ b/score/bitmanipulation/bit_manipulation_test.cpp
@@ -232,6 +232,23 @@ TEST(SetBit, WithUInt64)
     EXPECT_EQ(input, expectedResult);
 }
 
+TEST(SetBit, WithNegativeInt32)
+{
+    RecordProperty("PartiallyVerifies", "comp_req__bitmanipulation__bit_operations");
+    RecordProperty("Description",
+                   "Check that SetBit sets the requested bit of a negative (sign-bit-set) int32_t value, i.e. "
+                   "the internal unsigned promotion and narrowing back to a signed type do not corrupt the result.");
+    RecordProperty("TestType", "requirements-based");
+    RecordProperty("DerivationTechnique", "equivalence-classes");
+
+    std::int32_t input{-2};                     // 0xFFFFFFFE
+    constexpr std::int32_t expectedResult{-1};  // 0xFFFFFFFF
+
+    EXPECT_TRUE(::score::platform::SetBit(input, 0U));
+
+    EXPECT_EQ(input, expectedResult);
+}
+
 TEST(SetBit, OverflowWithUInt8)
 {
     RecordProperty("PartiallyVerifies", "comp_req__bitmanipulation__bounds_safety");
--- a/score/bitmanipulation/bit_manipulation_test.cpp
+++ b/score/bitmanipulation/bit_manipulation_test.cpp
@@ -236,14 +236,19 @@ TEST(SetBit, OverflowWithUInt8)
 {
     RecordProperty("PartiallyVerifies", "comp_req__bitmanipulation__bounds_safety");
     RecordProperty("Description",
-                   "Check that SetBit returns false when the bit position exceeds the width of a uint8_t value.");
+                   "Check that SetBit returns false and leaves the value unmodified when the bit position exceeds "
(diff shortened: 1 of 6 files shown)


score/bitmanipulation/bit_manipulation_test.cpp
```diff
@@ -151,6 +234,12 @@ TEST(SetBit, WithUInt64)
 
 TEST(SetBit, OverflowWithUInt8)
 {
+    RecordProperty("PartiallyVerifies", "comp_req__bitmanipulation__bounds_safety");
```
fbaeuerle, line 254: From the requirement:
>   The bit manipulation functions shall validate input parameters against bounds and, on out-of-bounds access, shall leave the target value unmodified and return false.

it is nowhere verified (in all tests linked to that requirement) that the target value remains unmodified

score/bitmanipulation/bitmask_operators_test.cpp
```diff
@@ -41,15 +42,13 @@ namespace
 {
 using UnderlyingType = std::underlying_type_t<MyBitmask>;
 
-TEST(MyBitmask, UnderlyingValuesMatchExpectations)
```
fbaeuerle, line 44: Is this test nonsense or is there no requirement for it?

fbaeuerle (changes requested):
linkage looks good, but there's a a testing gap (which now becomes obvious when reading the linked requirement)

score/bitmanipulation/bitmask_operators_test.cpp
4og: It is nonsense. The enum `MyBitmask` is defined on line 24 in this file

score/bitmanipulation/bit_manipulation_test.cpp
4og: Fixed

score/bitmanipulation/bit_manipulation_test.cpp
```diff
@@ -28,41 +28,78 @@ using ::testing::Eq;
 
 TEST(HalfByte, CanBeConstructedFromUInt8)
 {
+    RecordProperty("PartiallyVerifies", "comp_req__bitmanipulation__byte_operations");
```
fbaeuerle, line 31: found a mismatch between implementation and requirement:
>   The bit manipulation component shall provide an API for extracting and setting bytes and half-bytes for any integral type up to 64 bits, returning boolean success status.

it doesn't quite match the implementation, we have the byte and halfbyte types. we also have `GetByte`, which does not return bool (and i wonder how it could possibly return a failure). we do however, do not have a `SetByte` counterpart. and even on the `SetByte`counterpart i see little value in reporting success/failure as a boolean (all errors i can think of can be caught during compilation).

4og: I addressed this by splitting the requirement. The bool status is removed, and the  "setting bytes" is also removed.

fbaeuerle: https://github.com/eclipse-score/baselibs/blob/main/score/bitmanipulation/docs/requirements/chklst_req_inspection.rst?plain=1#L137

does this now need an update? the linked finding (https://github.com/eclipse-score/score/issues/2698) was already closed (i see back then you added those requirements)

fbaeuerle (approved):

score/bitmanipulation/bit_manipulation_test.cpp
4og: Eventually, we will need to do the re-inspection. I dropped the `:tags: inspected` to indicate that new requirement inspection is needed

Issue #512, opened 2026-08-25
Integrate coverage infrastructure provided by score_tooling


4og:
### What

The score_tooling provided the LLVM source-based coverage solution for both C++ and Rust, The goal of this task to integrate this solution to baselibs.

See adoption of the llvm-cov in `score_logging`: https://github.com/eclipse-score/logging/pull/282/

(See also the evaluation in #289)

### Acceptance Criteria (DoD)

- [ ] The LLVM-based coverage solution is integrated
- [ ] The currently used gcov-based implementation is removed from baselibs

### How

See https://github.com/eclipse-score/tooling/blob/main/coverage/README.md

PR #513, opened 2026-08-25
chore(bazel): bump score_tooling, score_crates and related deps

4og:
## Summary

- Update `score_docs_as_code`, `score_process_description`, and `score_platform` to versions whose archives don't depend on `api.github.com`, which is hard to reach from corporate infrastructure.
- Bump `score_tooling` from 1.3.1 to 2.2.1 and `score_crates` from a pinned `git_override` commit to the released 0.0.11.
- Bump `aspect_rules_lint` to 2.5.0 and `examples/integration`'s `score_bazel_platforms` to 1.0.0 to match.
- Unset `ANDROID_HOME` in `examples/integration/.bazelrc` to avoid Android SDK search failures on GitHub Actions once the runner's disk space is cleaned up.

## score_tooling 2.2.1 follow-up fixes

This score_tooling version made breaking API changes that broke the root `BUILD` file (and therefore `//:format.fix` and everything else in the root package):

- `use_format_targets` moved from `@score_tooling//:defs.bzl` to `@score_tooling//third_party/format:macros.bzl`.
- `rust_coverage_report` was removed upstream (old Ferrocene coverage flow). `//:rust_coverage` and `//:rust_coverage_report` are dropped for now. The `.github/workflows/rust_coverage.yml` reusable workflow (which drove the removed flow through `//:rust_coverage`) and the `ferrocene-coverage` `.bazelrc` config it relied on were removed as well, since that workflow was already broken. Migrating to the new LLVM-based coverage pipeline is tracked separately in eclipse-score/baselibs#512.
- `//:copyright-check` also broke: its `cr_checker.bzl` now forwards `srcs` verbatim to `git ls-files` as pathspecs, so the Bazel labels (`//:BUILD`, `//:MODULE.bazel`, `//:qemu.bzl`) previously in `BUILD`'s `srcs` list made `git ls-files` fail. Switched to plain workspace-relative paths, then ran `//:copyright-fix` to normalize the headers the rewritten checker now flags.


BUILD
```diff
@@ -181,21 +182,10 @@ dash_license_checker(
...
-
-alias(
-    name = "rust_coverage_report",
-    actual = ":rust_coverage",
-    visibility = ["//visibility:public"],
-)
+# TODO: rust_coverage_report was removed by score_tooling >= 2.1.0
+# //:rust_coverage and //:rust_coverage_report are gone until the repo migrates
+# to the new score_coverage_scope/score_coverage_reporter LLVM pipeline
+# https://github.com/eclipse-score/baselibs/issues/512
```
fbaeuerle, line 188: it is also still referenced in https://github.com/eclipse-score/baselibs/blob/main/.github/workflows/rust_coverage.yml#L25

fbaeuerle (changes requested):

BUILD
4og: I removed the workflow and the entries in bazelrc

fbaeuerle (approved):

Issue #496, opened 2026-08-20
[BUILD] acl target: prebuilt acl-deb is non-PIC → link failure on Ubuntu 24.04 / ARM64


tfa2si:
## Context

`score_baselibs` provides an `acl` target for POSIX ACL support (used by S-CORE's SHM access control). The upstream target is an `alias` that selects a prebuilt `.deb` archive:

```starlark
alias(
    name = "acl",
    actual = select({
        "@platforms//cpu:aarch64": "@acl-deb-aarch64//:acl",
        "//conditions:default": "@acl-deb//:acl",
    }),
    visibility = ["//:__subpackages__"],
)
```

## Problem

On Ubuntu 24.04 (and ARM64 cross-builds) we hit build/link issues with the prebuilt non-PIC static `.a` shipped in `acl-deb`. We work around it by overriding the `acl` target to link the system `libacl` instead:

```starlark
load("@rules_cc//cc:defs.bzl", "cc_library")

cc_library(
    name = "acl",
    linkopts = ["-lacl"],
    visibility = ["//:__subpackages__"],
)
```

## Question for maintainers

1. Is the prebuilt `acl-deb` archive known to be **non-PIC**? We observed linker failures when it is pulled into shared/position-independent link actions on Ubuntu 24.04 (`cannot be used when making a shared object; recompile with -fPIC` style).
2. Would it be acceptable/robust to make the `acl` target link against the **system `libacl`** (e.g. `-lacl` / `libacl1-dev`) as the default, rather than bundling the prebuilt `.deb`? That is how most Bazel modules resolve common system libraries and avoids the PIC problem entirely.
3. Alternatively, is the non-PIC issue specific to how the archive is produced, and would rebuilding `acl-deb` with `-fPIC` be the preferred upstream fix?

## Supporting info

- We currently carry this as a local patch applied at fetch time (`single_version_override` → `patches`) in `MODULE.bazel` — shown below, for reference/our own tracking (we are not in a position to contribute upstream).
- With the override, the full HFC ARM64 deployment builds and runs cleanly (15/15 tests), so the system-lib approach is proven in practice.

### Local patch (for reference — not a contribution offer)

```patch
--- a/third_party/acl/BUILD
+++ b/third_party/acl/BUILD
@@
-alias(
-    name = "acl",
-    actual = select({
-        "@platforms//cpu:aarch64": "@acl-deb-aarch64//:acl",
-        "//conditions:default": "@acl-deb//:acl",
-    }),
[...]

Issue #78, opened 2026-01-23
Baselibs doesn't work on AArch64 Linux due to libacl dependency being x86_64 only


4og:
Potential fix:

```
diff --git a/third_party/acl/BUILD b/third_party/acl/BUILD
index 7fdb429..0d6e743 100644
--- a/third_party/acl/BUILD
+++ b/third_party/acl/BUILD
@@ -1,5 +1,19 @@
+
+
+cc_library(
+    name = "libacl",
+    linkopts = [
+        "-lacl",
+    ],
+    target_compatible_with = ["@platforms//os:linux", "@platforms//cpu:aarch64"],
+    visibility = ["//visibility:public"],
+)
+
 alias(
     name = "acl",
-    actual = "@acl-deb//:acl",
+    actual = select({
+		"@platforms//cpu:aarch64": ":libacl",
+		"//conditions:default": "@acl-deb//:acl",
+		}),
     visibility = ["//visibility:public"], 
 )
\ No newline at end of file

```

_Originally posted by @btirunagaru in https://github.com/eclipse-score/score/discussions/2462#discussioncomment-15575110_

PR #515, opened 2026-08-26
libacl import options to address -fPIC and other similar issue

Nikhil2206:
This is Just draft PR to compare and finalized the approach to import libacl in baselib repo

### Replace prebuilt acl `.deb` with a hermetic, source-built `libacl`

**Problem:** `score_baselibs` consumed `acl` as an architecture-pinned Ubuntu `.deb` (`libacl1-dev`). This repeatedly broke on non-Ubuntu targets and on `aarch64` (missing `-fPIC`, arch/toolchain mismatches) — tracked as SWP-278650 (#172, #496, #78).

**Fix:** Vendor upstream `acl-2.4.0` source and compile it with this repo's own toolchain instead, so PIC/ABI can never drift from the rest of the build, on any target this repo supports (Ubuntu/AutoSD/Elektrobit, x86_64/aarch64) — no per-distro `.deb` needed.

### What's included

- **`MODULE.bazel`**: replaced the two `deb()` rules (`acl-deb`, `acl-deb-aarch64`) with a single `archive()` rule (`acl-src`) pulling the sha256-pinned `acl-2.4.0` tarball.
- **`acl.BUILD`** / **`acl_sources.bzl`**: `cc_library(:acl)` compiling only the plain POSIX ACL subset actually needed by `acl_impl.cpp` (excludes the GPL-2 CLI tools and the unused `_at`/`perm_copy_*` files).
- **`config.h`**: hand-written stand-in for acl's autoconf-generated `config.h`, scoped to this repo's one supported target class (glibc, 64-bit, little-endian Linux).
  - `ENABLE_NLS` is the one genuine configure-time option (gates `gettext()` in `include/misc.h`), so it's exposed as a real Bazel flag (`bool_flag` + `config_setting`, `--@score_baselibs//third_party/acl:enable_nls`) instead of being hardcoded.
- **`check_config_drift.py`** (`config_drift_test`): fails the build if (a) a vendored source references an autoconf-style macro (`HAVE_*`, `ENABLE_*`, ...) not accounted for in `config.h`, or (b) a vendored file's header claims the GPL rather than the LGPL — both catch drift automatically on future `acl` version bumps.
- **`cc_shared_library(:acl_shared)`**: builds `libacl` as a real shared object, giving consumers a straightforward LGPL-2.1 §6(b) compliance path (swap the `.so` without relinking). Static linking (`:acl`) is now visibility-restricted to the one wrapper that needs it at compile time (`score/os:acl`); every other consumer must go through `:acl_shared` + `dynamic_deps`. `score/os/test:acl_test` and `score/os/utils/acl:unit_test` were switched to dynamic linking as the first (and currently only) consumers.
- **`NOTICE`** / **`acl-lgpl.txt`**: added LGPL-2.1 attribution and license text for the vendored code.
- **`.bazelrc`**: demoted `-Werror` for `external/.*acl-src.*`, since the vendored upstream source doesn't meet this repo's strict warning bar (e.g. `-Wcast-qual`).
- **`libacl_import_options.md`**: design doc comparing this approach (build-from-source) against a hermetic sysroot alternative, grounded in the actual implementation.

### Validation
Built and tested on both `bl-x86_64-linux` and `bl-aarch64-linux` (aarch64 via qemu); verified via `ldd`/runfiles inspection that the dynamic-linked test targets pull in `libacl_shared.so` at runtime with no static duplication; confirmed the visibility restriction blocks unauthorized static consumers and `config_drift_test`/GPL-check both pass.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -41,6 +41,9 @@ build --credential_helper="*.qnx.com=%workspace%/.github/tools/qnx_credential_he
 build --features=minimal_warnings --features=-strict_warnings --features=warnings_as_errors
 build --incompatible_strict_action_env
 build --per_file_copt=external/.*google_benchmark.*@-Wno-error
+# acl is vendored third-party source (see third_party/acl/acl.BUILD); it does not
+# meet this repo's strict warning bar (e.g. -Wcast-qual), so demote its warnings.
+build --per_file_copt=external/.*acl-src.*@-Wno-error
 
 # Common flags for tests
 test:_bl_common --build_tests_only
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -160,18 +160,20 @@ llvm.toolchain(
 use_repo(llvm, "llvm_toolchain")
 
 deb = use_repo_rule("@download_utils//download/deb:defs.bzl", "download_deb")
+archive = use_repo_rule("@download_utils//download/archive:defs.bzl", "download_archive")
 
-deb(
-    name = "acl-deb",
+# acl is vendored from upstream source instead of a prebuilt, arch-pinned Ubuntu .deb
+# (see third_party/acl/acl.BUILD and score/os/BUILD's `:acl` target) to avoid the
+# recurring -fPIC/arch breakage tracked by SWP-278650 (eclipse-score/baselibs#172, #496, #78).
+archive(
+    name = "acl-src",
     build = "//third_party/acl:acl.BUILD",
-    urls = ["https://archive.ubuntu.com/ubuntu/pool/main/a/acl/libacl1-dev_2.3.1-1_amd64.deb"],
-    visibility = ["//visibility:public"],
-)
-
-deb(
-    name = "acl-deb-aarch64",
-    build = "//third_party/acl:acl.BUILD",
-    urls = ["https://ports.ubuntu.com/ubuntu-ports/pool/main/a/acl/libacl1-dev_2.3.1-1_arm64.deb"],
+    integrity = "sha256-5mETFFbScIoBxhSg9ADhHX0b+utvPnS3W7mAty8BYaM=",
+    strip_prefix = "acl-2.4.0",
+    urls = [
+        "https://download.savannah.nongnu.org/releases/acl/acl-2.4.0.tar.xz",
+        "http://download-mirror.savannah.nongnu.org/releases/acl/acl-2.4.0.tar.xz",
+    ],
     visibility = ["//visibility:public"],
 )
 
--- a/NOTICE
+++ b/NOTICE
@@ -54,3 +54,16 @@ The license text can be found in licenses/llvm-libcxx.txt.
 //===----------------------------------------------------------------------===//
 
 [1] https://github.com/llvm/llvm-project/tree/main/libcxx
+
+## acl
+
+third_party/acl vendors the plain POSIX ACL portion of libacl's source (excluding
+acl_delete_def_file_at and perm_copy_fd/perm_copy_file) from acl 2.4.0 [2],
+compiled directly as a score_baselibs cc_library instead of installing a prebuilt,
+architecture-pinned Ubuntu .deb package (see SWP-278650 and
+https://github.com/eclipse-score/baselibs/issues/172, /issues/496, /issues/78).
+
+libacl is provided under the GNU Lesser General Public License v2.1 (or later).
+The license text can be found in licenses/acl-lgpl.txt.
+
+[2] https://download.savannah.nongnu.org/releases/acl/acl-2.4.0.tar.xz
--- a/third_party/acl/BUILD
+++ b/third_party/acl/BUILD
@@ -11,11 +11,57 @@
 # SPDX-License-Identifier: Apache-2.0
 # *******************************************************************************
 
+load("@bazel_skylib//rules:common_settings.bzl", "bool_flag")
+load("@rules_cc//cc:defs.bzl", "cc_library")
+
+# Bazel-native equivalent of upstream acl's `./configure --enable-nls`: gates
+# whether include/misc.h's `_()` macro (see @acl-src//:acl) routes through gettext().
+bool_flag(
+    name = "enable_nls",
+    build_setting_default = False,
+)
+
+config_setting(
+    name = "nls_enabled",
+    flag_values = {":enable_nls": "True"},
+    visibility = ["@acl-src//:__pkg__"],
+)
+
+# Hand-written stand-in for acl's autoconf-generated include/config.h, consumed by
+# @acl-src (see MODULE.bazel and acl.BUILD) which vendors libacl from source.
+cc_library(
+    name = "config_h",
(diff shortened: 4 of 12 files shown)


third_party/acl/libacl_import_options.md
```diff
@@ -0,0 +1,68 @@
...
+
+- **Pros:** Fully hermetic, arch-agnostic (any target the toolchain supports),
+  immune to distro packaging quirks (no more "was this `.deb` built with
+  `-fPIC`" issues); consistent story across Ubuntu/AutoSD/Elektrobit since none
+  of them are involved at build time.
+- **Cons:** Higher upfront effort (write BUILD files for each source tree,
+  handle their native build systems/autoconf quirks); becomes something this
+  repo now owns and must patch/update over time; still need libc/system
+  headers (e.g. `sys/capability.h`) to match target consistently.
+
```
4og, line 118: Regarding the libcap. There is https://registry.bazel.build/modules/libcap/2.27.bcr.2/overview. Can we use it instead of pulling a deb? I'm not entirely sure it's the same thing or it's a different implementation

Maybe we can also add the acl lib to BCR?

third_party/acl/libacl_import_options.md
```diff
@@ -0,0 +1,68 @@
...
+- **Pros:** Fully hermetic and reproducible; single point of version control
+  (the sysroot artifact); naturally extends the toolchain infrastructure
+  already in place for GCC/QCC.
+- **Cons:** Requires building/maintaining that sysroot artifact (likely
+  upstream in `score_bazel_cpp_toolchains` or a new repo) — work isn't only in
+  this repo; slower to iterate; still needs a strategy per distro if
+  AutoSD/Elektrobit need different library builds (e.g. musl vs glibc,
+  different ABI).
+
```
4og, line 88: Another con here:
- users of baselibs may use their own toolchain that could be lacking the appropriate sysroot configuration

third_party/acl/libacl_import_options.md
4og: I looked again. The BCR has the same libcap we're using, but a very old version. We can raise a PR in BCR to add a newer version

third_party/acl/libacl_import_options.md
Nikhil2206: agree

third_party/acl/libacl_import_options.md
Nikhil2206: we could think of BCR approach, but this might look overkill for baselib purpose because 
- we likely need to maintain that per lib, 
- could have delayed merges, maintainer may or may not like this
-  harder to audit for a safety case, and re-verification is needed on every module version bump, on someone else's release cadence.

i think keeping build and config controll to baselib look relatively safer

Nikhil2206 (commented):
@nradakovic 
could you please share your openion on this topic, 
specifically option 4 using build from source approach, we are likely to use,  I am currently verifying it in reference integration

third_party/acl/BUILD
```diff
@@ -11,11 +11,57 @@
 # SPDX-License-Identifier: Apache-2.0
 # *******************************************************************************
 
+load("@bazel_skylib//rules:common_settings.bzl", "bool_flag")
+load("@rules_cc//cc:defs.bzl", "cc_library")
+
+# Bazel-native equivalent of upstream acl's `./configure --enable-nls`: gates
+# whether include/misc.h's `_()` macro (see @acl-src//:acl) routes through gettext().
+bool_flag(
```
4og, line 19: Why is this flag needed? Do we really want to give a user a way to control it?

.bazelrc
```diff
@@ -41,6 +41,9 @@ build --credential_helper="*.qnx.com=%workspace%/.github/tools/qnx_credential_he
 build --features=minimal_warnings --features=-strict_warnings --features=warnings_as_errors
 build --incompatible_strict_action_env
 build --per_file_copt=external/.*google_benchmark.*@-Wno-error
+# acl is vendored third-party source (see third_party/acl/acl.BUILD); it does not
+# meet this repo's strict warning bar (e.g. -Wcast-qual), so demote its warnings.
+build --per_file_copt=external/.*acl-src.*@-Wno-error
```
4og, line 46: We should find another way to solve this as acl is non-dev-dependency. @nradakovic, any options here other that creating a yet another transition?

third_party/acl/BUILD
```diff
@@ -11,11 +11,57 @@
...
+# target that needs the static compile-time interface, everyone else must use
+# :acl_shared (paired with dynamic_deps) instead.
 alias(
     name = "acl",
-    actual = select({
-        "@platforms//cpu:aarch64": "@acl-deb-aarch64//:acl",
-        "//conditions:default": "@acl-deb//:acl",
-    }),
+    actual = "@acl-src//:acl",
+    visibility = ["@score_baselibs//score/os:__pkg__"],
```
4og, line 60: These two aliases looks wrong.
The score/os is the only user of libacl and here it links statically to it. What is purpose of `acl_shared` then? 

My expectation nobody is linking to the static lib other then for purposes of constructing a shared library. I.e. score/os users should link to the shared lib only

MODULE.bazel
```diff
@@ -139,18 +160,20 @@ llvm.toolchain(
...
-    build = "//third_party/acl:acl.BUILD",
-    urls = ["https://archive.ubuntu.com/ubuntu/pool/main/a/acl/libacl1-dev_2.3.1-1_amd64.deb"],
-    visibility = ["//visibility:public"],
-)
-
-deb(
-    name = "acl-deb-aarch64",
+# acl is vendored from upstream source instead of a prebuilt, arch-pinned Ubuntu .deb
+# (see third_party/acl/acl.BUILD and score/os/BUILD's `:acl` target) to avoid the
+# recurring -fPIC/arch breakage tracked by SWP-278650 (eclipse-score/baselibs#172, #496, #78).
```
4og, line 167: Avoid mentioning any internal tickets in code.

NOTICE
```diff
@@ -54,3 +54,16 @@ The license text can be found in licenses/llvm-libcxx.txt.
 //===----------------------------------------------------------------------===//
 
 [1] https://github.com/llvm/llvm-project/tree/main/libcxx
+
+## acl
+
+third_party/acl vendors the plain POSIX ACL portion of libacl's source (excluding
+acl_delete_def_file_at and perm_copy_fd/perm_copy_file) from acl 2.4.0 [2],
+compiled directly as a score_baselibs cc_library instead of installing a prebuilt,
```
4og, line 62: Mentioning ubuntu deb packages looks wrong here

4og (changes requested):

MODULE.bazel
Nikhil2206: I was trying it but, it got missed somehow

Nikhil2206:
@4og sure I will clean up the nitpicking things mentioned altogether once we reached decision

PR #520, opened 2026-08-26
Add vajson as second serializer backend

JuriSchroeder:
Adds vajson as a second json serialization backend. It's currently selectable by a CMake switch, but could replace the json_serializer implementation in the future. (See also https://github.com/orgs/eclipse-score/discussions/2390#discussioncomment-18160655)


4og:
@mihajlo-k, can you please review this PR?

score/json/internal/writer/vajson/writer/serializers/structures/generic_value_serializer.h
```diff
@@ -0,0 +1,253 @@
...
+    template <typename T>
+    // coverity[autosar_cpp14_m9_3_3_violation]
+    auto operator<<(JNumberType<T> number) && noexcept -> Next
+    {
+        return this->Serialize([this, number]() noexcept {
+            // Buffer size: max 24 chars for double, ~20 for int64, extra space for safety
+            std::array<char, 64> buffer{};
+            T value = static_cast<T>(number.GetValue());
+
+            const auto conversion_result = std::to_chars(buffer.data(), buffer.data() + buffer.size(), value);
```
mihajlo-k, line 119: According to RFC8259: https://www.rfc-editor.org/info/rfc8259/#section-6, infinities, NaNs or similar are not permitted. However, we don't guard against those values here. If we were to pass e.g. `std::numeric_limits<double>::infinity()` to the `std::to_chars` it will write (probably) a string "inf" into the buffer.

maybe guard it with `std::isfinite`?

score/json/internal/writer/vajson/writer/serializers/util/escaped_json_string.h
```diff
@@ -0,0 +1,130 @@
...
+/// - Otherwise:
+/// - Serialize the character directly.
+/// \param[in] os Output stream to write into.
+/// \param[in] string to escape and serialize.
+/// \return A reference to the output stream.
+auto inline operator<<(std::ostream& os, EscapedJsonString string) noexcept -> std::ostream&
+{
+    for (const char ch : string.GetValue())
+    {
+        switch (std::char_traits<char>::to_int_type(ch))
```
mihajlo-k, line 79: According to RFC8259: https://www.rfc-editor.org/info/rfc8259/#section-7 we should also escape control characters (U+0000 through U+001F).

score/json/internal/writer/vajson/writer/serializers/stl/associative_containers.h
```diff
@@ -0,0 +1,155 @@
...
+/// \tparam Alloc Type of allocator.
+/// \param[in] serializer instance to write into.
+/// \param[in] map Map to serialize.
+/// \return The succeeding serializer.
+template <typename Next, typename Key, typename Value, typename Cmp, typename Alloc>
+auto operator<<(GenericValueSerializer<Next>&& serializer, const std::map<Key, Value, Cmp, Alloc>& map) noexcept ->
+    typename GenericValueSerializer<Next>::Next
+{
+    return std::move(serializer) << JObject(map);
+}
```
mihajlo-k, line 97: duplicate. please remove one

mihajlo-k (changes requested):
I left a couple of comments. please note that this is a partial review, and I might post more comments.

score/json/internal/writer/vajson/writer/serializers/stl/associative_containers.h
mihajlo-k: so, actually this should not compile. and now when I look at it, it seems it is indeed not compiled anywhere in the code. it's only included in `score/json/internal/writer/vajson/writer/serializers.h` but `serializers.h` itself is not included anywhere. @JuriSchroeder please check if this chunk of code is needed at all, and remove if not.

score/json/internal/writer/vajson/writer/types/bin_types.h
mihajlo-k, line 1: Are these types used in `VajsonSerialize`? are they even related to JSON serialization?

score/json/internal/writer/vajson/vajson_serialize_test.cpp
mihajlo-k, line 1: Don't forget to add tests metadata to the tests (`RecordProperty(...)`). see some existing tests for this

score/json/internal/writer/vajson/writer/serializers/structures/generic_value_serializer.h
```diff
@@ -0,0 +1,253 @@
...
+    /// \param[in] number value to serialize.
+    /// \return The succeeding serializer.
+    template <typename T>
+    // coverity[autosar_cpp14_m9_3_3_violation]
+    auto operator<<(JNumberType<T> number) && noexcept -> Next
+    {
+        return this->Serialize([this, number]() noexcept {
+            // Buffer size: max 24 chars for double, ~20 for int64, extra space for safety
+            std::array<char, 64> buffer{};
+            T value = static_cast<T>(number.GetValue());
```
mihajlo-k, line 117: what is the point of `JNumberType`? to me it seemed that it is "widening" the underlying type, but in this line it's narrowed back again.

score/json/internal/writer/vajson/vajson_serialize.h
```diff
@@ -0,0 +1,200 @@
...
+        serialized.emplace(std::move(serializer) << score::json::vajson::JNull());
+    }
+    else if (const auto number = value.As<score::json::Number>(); number.has_value())
+    {
+        serialized.emplace(SerializeNumber(std::move(serializer), number->get()));
+    }
+    else
+    {
+        const auto boolean = value.As<bool>();
+        serialized.emplace(std::move(serializer) << score::json::vajson::JBool(*boolean));
```
mihajlo-k, line 144: I know we currently exhaust all other possibilities before we go to `bool`, but still I think that we shouldn't simply dereference a `Result` type before checking it's validity.

score/json/internal/writer/vajson/vajson_serialize.cpp
```diff
@@ -0,0 +1,79 @@
...
+    auto serializer = score::json::vajson::DocumentSerializer{std::ref(out_stream)};
+    static_cast<void>(std::move(serializer) << json_data);
+
+    // A default constructed Result<void> already carries the success state.
+    score::Result<void> result{};
+    if (out_stream.fail())
+    {
+        result = score::Result<void>{
+            score::unexpect,
+            score::json::MakeError(score::json::Error::kUnknownError, "vaJSON serializer failed to write to stream")};
```
mihajlo-k, line 32: this can be simplified
```suggestion
        result = MakeUnexpected(score::json::Error::kUnknownError, "vaJSON serializer failed to write to stream");
```

score/json/README.md
```diff
@@ -378,3 +380,15 @@ In order to make use of nlohmann json library, feature flag needs to be set. Ple
...
 
 nlohmann json library do not supports hexadecimal. As it is not part of json standard.
+
+## Selecting writer library
+
+Independently of the parser, the serialization backend is selected by the `writer_library` flag. It accepts
+`json_serialize` (the default, a custom implementation) and `vajson` (the vector json library). The parser flag
+`base_library` has no influence on serialization.
+
+bazel test --config=spp_host_clang //score/json/... --//platform/aas/lib/json:writer_library="vajson"
```
mihajlo-k, line 390:
```suggestion
bazel test --config=spp_host_clang //score/json/... --//score/json:writer_library="vajson"
```

mihajlo-k (changes requested):
Generally a very good addition with a lot of improvements to existing code. However, there might be a couple of unnecessary additions, and a few potential improvements.

Additionally, please remove any commits that are out of this scope from the PR. e.g. [Remove amsr namespace; Remove VCA annotations](https://github.com/eclipse-score/baselibs/pull/520/changes/755faa95d8a1232a5aa0bb5cd589d3305128816a)

score/json/internal/writer/vajson/writer/serializers/structures/generic_value_serializer.h
JuriSchroeder: I added a check for that.

score/json/internal/writer/vajson/writer/serializers/util/escaped_json_string.h
JuriSchroeder: Fixed.

score/json/internal/writer/vajson/writer/serializers/stl/associative_containers.h
JuriSchroeder: True. This was dead code because the adapter I built from S-Core to vaJSon, never actually uses these types. There was some more dead code I removed as well.

score/json/internal/writer/vajson/writer/types/bin_types.h
JuriSchroeder: Removed. (This was related to an extension, but since binary parsing is also removed it should also be removed from the serializer.)

score/json/internal/writer/vajson/vajson_serialize.h
JuriSchroeder: Done

score/json/internal/writer/vajson/vajson_serialize.cpp
JuriSchroeder: Done

score/json/README.md
JuriSchroeder: Done

score/json/internal/writer/json_serialize/json_serialize_backend.cpp
```diff
@@ -0,0 +1,262 @@
...
+#include <sstream>
+#include <type_traits>
+
+namespace
+{
+
+template <class U>
+// Coverity thinks this function is unused, but it is used in for calculating kIntBufLen.
+// coverity[autosar_cpp14_a0_1_3_violation]
+constexpr std::size_t max_dec_digits() noexcept
```
sankurm: Unnecessary implementation: We don't need this as the standard has this as a `constexpr`.
See [std::numeric_limits<T>::digits10](https://cppreference.com/cpp/types/numeric_limits/digits10)
In my opinion, this can be removed.

score/json/internal/writer/vajson/writer/serializers/structures/generic_value_serializer.h
JuriSchroeder: The JNumberType is used to force correct serialization of different number types, e.g. , to prevent printing "A" for os << std::uint8_t{65} instead of "65". But in practice this is irrelevant now because the adapter in vajson_serialize.h only uses the JNumberType instantiations for std::uint64_t, std::int64_t, float and double.
I removed the code related to the unused types.

mihajlo-k (approved):

4og (approved):

PR #526, opened 2026-08-27
enable contributions for score/**/docs/**

Rutik7:


.github/workflows/restricted_paths.yml
```diff
@@ -40,6 +40,7 @@ jobs:
               - 'score/mw/**'
               - 'score/memory/shared/**'
               - 'score/memory/design/shared_memory/**'
+              - '!score/**/docs/**'
```
4og, line 42: For this to work, the `predicate-quantifier: some-with-excludes` needs to be added above.
See https://github.com/dorny/paths-filter/blob/master/README.md

.github/workflows/restricted_paths.yml
```diff
@@ -40,6 +40,7 @@ jobs:
               - 'score/mw/**'
               - 'score/memory/shared/**'
               - 'score/memory/design/shared_memory/**'
```
4og: Can you please also unrestrict memory?

4og (changes requested):

.github/workflows/restricted_paths.yml
Rutik7: resolved

.github/workflows/restricted_paths.yml
Rutik7: Done

4og (approved):

4og (approved):

Issue #522, opened 2026-08-26
Clean up test-to-requirement traceability across baselibs components


4og:
## Goal

Bring every baselibs component's test-to-requirement traceability to a state that actually supports the ISO 26262 / ASIL-B verification evidence the codebase relies on: every `comp_req` is verified by tests that genuinely exercise it, every test that claims to verify a requirement actually does, and the requirement text itself is accurate.

## Background

This started with the bitmanipulation component (#499), where linking its existing tests to their `comp_req` needs surfaced more than missing links: a bounds-safety requirement whose tests never asserted the "leave the value unmodified" behavior it mandates, and a requirement that claimed an unimplemented "set byte" capability with a return-type convention that didn't match the code. Both were fixed, along with splitting an overly broad requirement into two atomic ones. That work is the template for the tasks linked under this Epic.

## Scope

Each linked task scopes this same clean-up to one or a few components. The result should close genuine gaps without padding out test combinations that already share the same implementation path, and should reconcile with any context that already exists for that component (open PR review threads, architecture docs, requirements inspection checklists) rather than duplicating or contradicting it.

Example of a prompt to help with the tasks:

```
/requirements-management-skill Bring the <component> component's requirement-to-test traceability
(C++ gtest and Rust unit tests anywhere under score/<component>/) to a state where:
 - Every comp_req's normative text accurately reflects what's implemented: no requirement claims a
  capability that doesn't exist, describes API mechanics instead of observable behavior, or bundles
  multiple, independently testable capabilities under one ID that would be more atomic split apart.
 - Any scope a requirement claims is actually reflected by
  the tests covering it, in both directions: neither claimed-but-untested nor tested-but-unclaimed.
 - Every requirement's coverage matches what's appropriate for its nature: functional requirements have
  real test coverage, while structural/non-functional requirements correctly verified by inspection
  are left alone rather than forced into a test that can't meaningfully exist.
 - Every test either verifies a real comp_req or has a clear, deliberate reason not to.
 - The result is neither under- nor over-tested: any change closes a genuine gap in observable behavior
  rather than padding out combinations that already share the same implementation path.
 - Nothing here contradicts or duplicates context that already exists: open PR review threads on this
  component, its architecture doc, and its requirements inspection checklist, if any.

Report the current state against that bar, and check with me before any change beyond mechanical
metadata cleanup (rewording, splitting, or authoring a requirement; removing a test). Have an
independent sub-agent sanity-check the result against the skill's own linkage-review criteria.
```

PR #528, opened 2026-08-28
serialization: add EnumTraits<T> for named enum serialization

kevmeyer:
## Summary
Adds `score::common::visitor::EnumTraits<T>` — a compile-time trait template that lets an enum type register its enumerator name<->value table — plus a new `payload_tags::enum_le` tag and `enum_serialized_descriptor`. `visit_as` now SFINAE-dispatches to the enum-aware descriptor when `EnumTraits<T>::kHasNames` is true, and falls back to the previous integer-only `memcpy_serialized_descriptor` path otherwise (default `kHasNames = false`, so any enum without a registered trait is unaffected).

The **wire format (serialized byte layout) is unchanged** — `enum_serialized_descriptor` uses the same `memcpy_serialized<sizeof(T)>` payload type as before, only exposing extra compile-time metadata (`wire_tag`, `enum_type`) for downstream consumers.

This lets downstream code generators (e.g. FIBEX generation for DLT/CANape tooling) recover proto enumerator names for enum-typed fields, which are otherwise lost once the enum is folded into a plain integral wire type.

Also fixes a latent signed/unsigned dispatch bug introduced by the new enum overloads: `std::is_signed<T>` is only meaningful for arithmetic types and is always `false` for enum types (enums are not arithmetic types per the standard), even when the enum's underlying type is signed. Added `details::is_signed_enum<T>`, which inspects `std::underlying_type_t<T>` and is only instantiated for enum types, and use it instead of `std::is_signed<T>` for the enum dispatch — otherwise every enum would incorrectly get `wire_tag = unsigned_le`.

## Testing
Added `serializer_visitor.named_enum_wire_format_unchanged` unit test in `test_serializer_visitor.cpp`:
- Registers `EnumTraits<test::EnumWithNames>` and verifies `kHasNames` defaults to `false` for enums without a trait (e.g. the pre-existing `test::E`).
- Verifies `check_serialized<test::EnumWithNames>() == sizeof(test::EnumWithNames)` (wire format unchanged).
- Static-asserts the `visit_as` descriptor for a named enum uses `payload_tags::enum_le`, exposes the correct `enum_type`, and dispatches to `wire_tag == payload_tags::signed_le` for an enum with a signed underlying type (regression check for the signed/unsigned dispatch fix).
- Verifies the registered enumerator name<->value table (`kEnumerators`) is actually retrievable through the trait.

`bazel test //score/static_reflection_with_serialization/serialization:unit_tests` — all pass (`serializer_ut`, `size_visitor_ut`, `visitor_type_traits_ut`).

### DIFF
--- a/score/static_reflection_with_serialization/serialization/include/serialization/visit_serialize.h
+++ b/score/static_reflection_with_serialization/serialization/include/serialization/visit_serialize.h
@@ -80,11 +80,65 @@ struct vector
 struct optional
 {
 };
+// Distinct tag for enum-typed fields whose enumerator name table is known via a specialization
+// of ::score::common::visitor::EnumTraits<T>. Carries the same wire-level integer encoding as
+// signed_le / unsigned_le (see enum_serialized_descriptor::wire_tag) but lets downstream
+// consumers (e.g. FIBEX generation) detect "this is a named enum" instead of falling through to
+// the generic integral overloads.
+struct enum_le
+{
+};
 
 }  // namespace payload_tags
 
+// One value <-> name mapping of an enumerator, as declared in the originating proto `enum`.
+struct Enumerator
+{
+    std::int64_t value;
+    const char* name;
+};
+
+// Per-enum-type metadata carrier, used by the serialization visitor and, downstream, by FIBEX
+// generation to recover the enumerator name<->value table that would otherwise be lost once the
+// enum is folded into a plain integral wire type.
+//
+// The primary (unspecialized) template intentionally carries no names: any C++ enum works with
+// the existing signed/unsigned integral serialization path unless a specialization of this trait
+// is provided for it. Code generators are expected to emit a specialization for every proto
+// `enum` they generate, reusing the enumerator list they already have in memory during codegen.
+// This is a compile-time-only, zero-runtime-cost side channel: it does not alter the wire format
+// of the enum itself.
+template <typename T>
+struct EnumTraits
+{
+    static constexpr bool kHasNames = false;
+};
+
 namespace details
 {
+/// \brief Compile-time predicate for "is T an enum with a signed underlying type?".
+///
+/// std::is_signed<T> is only meaningful for arithmetic types and is always false for enum types
+/// (enums are not arithmetic types per the standard), even when their underlying type is signed.
+/// This trait instead inspects the enum's underlying type, and is only instantiated for enum
+/// types (via the bool non-type template parameter) so it stays safe to use unconditionally in
+/// enable_if_t expressions alongside is_enum<T> for non-enum T.
+/// \tparam T Candidate type; only its is_enum-ness is inspected here (this primary/false-type
+///           template is selected for all non-enum T).
+template <typename T, bool = std::is_enum<T>::value>
+struct is_signed_enum : std::false_type
+{
+};
+
+/// \brief Specialization selected for enum types: inherits from
+///        std::is_signed<std::underlying_type_t<T>>, i.e. ::value is true iff T's underlying
+///        integer type is signed.
+/// \tparam T Enum type to inspect.
+template <typename T>
+struct is_signed_enum<T, true> : std::is_signed<typename std::underlying_type<T>::type>
+{
+};
+
 template <typename S, typename T>
 auto cast_to_source_serializable_data_span(const T* data, size_t size) -> score::cpp::span<const S>
 {
@@ -972,6 +1026,21 @@ struct memcpy_serialized_descriptor
     using payload_type = memcpy_serialized<sizeof(T)>;
 };
 
+// Descriptor for enum-typed fields whose enumerator names are known via EnumTraits<T>. Keeps the
+// same on-wire integer representation as memcpy_serialized_descriptor (Tag is signed_le or
+// unsigned_le, matching the enum's underlying signedness) but exposes the original enum type so
+// that downstream consumers (e.g. the FIBEX generator's visitor) can look up EnumTraits<T> and
+// recover the enumerator name<->value table. This does not change the serialized byte layout.
+template <typename Tag, typename T>
+// coverity[autosar_cpp14_a11_0_2_violation]
+struct enum_serialized_descriptor
+{
+    using payload_tag = payload_tags::enum_le;
+    using payload_type = memcpy_serialized<sizeof(T)>;
+    using wire_tag = Tag;
+    using enum_type = T;
+};
+
 template <typename A,
           typename T,
           std::enable_if_t<(std::is_integral<T>::value) && (std::is_signed<T>::value), std::int32_t> = 0>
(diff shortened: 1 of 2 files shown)


4og:
@rmaddikery, @hoppe-and-dreams, please check this PR

score/static_reflection_with_serialization/serialization/include/serialization/visit_serialize.h
```diff
@@ -80,11 +80,65 @@ struct vector
...
+// signed_le / unsigned_le (see enum_serialized_descriptor::wire_tag) but lets downstream
+// consumers (e.g. FIBEX generation) detect "this is a named enum" instead of falling through to
+// the generic integral overloads.
+struct enum_le
+{
+};
 
 }  // namespace payload_tags
 
+// One value <-> name mapping of an enumerator, as declared in the originating proto `enum`.
```
hoppe-and-dreams, line 94: This comment is misleading. This library is independent from proto. This might be you use case, but it is generally not correct to use this as the only use case of this library.

score/static_reflection_with_serialization/serialization/include/serialization/visit_serialize.h
```diff
@@ -80,11 +80,65 @@ struct vector
...
+    const char* name;
+};
+
+// Per-enum-type metadata carrier, used by the serialization visitor and, downstream, by FIBEX
+// generation to recover the enumerator name<->value table that would otherwise be lost once the
+// enum is folded into a plain integral wire type.
+//
+// The primary (unspecialized) template intentionally carries no names: any C++ enum works with
+// the existing signed/unsigned integral serialization path unless a specialization of this trait
+// is provided for it. Code generators are expected to emit a specialization for every proto
```
hoppe-and-dreams, line 107: Again, the whole proto description is not relevant for this library in S-CORE.

The general comment, about the difference between how "specialized" enums vs regular enums are handle suffices.

score/static_reflection_with_serialization/serialization/include/serialization/visit_serialize.h
```diff
@@ -972,6 +1026,21 @@ struct memcpy_serialized_descriptor
     using payload_type = memcpy_serialized<sizeof(T)>;
 };
 
+// Descriptor for enum-typed fields whose enumerator names are known via EnumTraits<T>. Keeps the
+// same on-wire integer representation as memcpy_serialized_descriptor (Tag is signed_le or
+// unsigned_le, matching the enum's underlying signedness) but exposes the original enum type so
+// that downstream consumers (e.g. the FIBEX generator's visitor) can look up EnumTraits<T> and
+// recover the enumerator name<->value table. This does not change the serialized byte layout.
+template <typename Tag, typename T>
+// coverity[autosar_cpp14_a11_0_2_violation]
```
hoppe-and-dreams, line 1035: No coverity in S-CORE repository

score/static_reflection_with_serialization/serialization/include/serialization/visit_serialize.h
```diff
@@ -1003,9 +1072,34 @@ inline auto visit_as(serialized_visitor<A>& /*unused*/, T& /*unused*/)
...
 
-template <typename A,
-          typename T,
-          std::enable_if_t<(std::is_enum<T>::value) && (std::is_signed<T>::value), std::int32_t> = 0>
+template <
+    typename A,
+    typename T,
+    std::enable_if_t<(std::is_enum<T>::value) && (details::is_signed_enum<T>::value) && (EnumTraits<T>::kHasNames),
+                     std::int32_t> = 0>
+/// \brief Dispatches a signed enum with a registered EnumTraits<T> specialization to the
```
hoppe-and-dreams, line 1080: Usually the comments go before the `template< ... >` line

score/static_reflection_with_serialization/serialization/include/serialization/visit_serialize.h
```diff
@@ -1003,9 +1072,34 @@ inline auto visit_as(serialized_visitor<A>& /*unused*/, T& /*unused*/)
...
+///        enum-aware descriptor.
+/// \param T Enum type with a signed underlying type and EnumTraits<T>::kHasNames == true.
+/// \return enum_serialized_descriptor<payload_tags::signed_le, T>, i.e. the same on-wire
+///         signed integer representation as memcpy_serialized_descriptor, but additionally
+///         exposing T so that EnumTraits<T> can be looked up by downstream consumers.
+// This is false positive, Overload signatures are different.
+// coverity[autosar_cpp14_m3_2_3_violation : FALSE]
+// coverity[autosar_cpp14_a2_10_4_violation : FALSE]
```
hoppe-and-dreams, line 1088: Again, coverity

score/static_reflection_with_serialization/serialization/include/serialization/visit_serialize.h
```diff
@@ -1014,9 +1108,34 @@ inline auto visit_as(serialized_visitor<A>& /*unused*/, T& /*unused*/)
...
+    std::enable_if_t<(std::is_enum<T>::value) && (!details::is_signed_enum<T>::value) && (EnumTraits<T>::kHasNames),
+                     std::int32_t> = 0>
+/// \brief Dispatches an unsigned enum with a registered EnumTraits<T> specialization to the
+///        enum-aware descriptor.
+/// \param T Enum type with an unsigned underlying type and EnumTraits<T>::kHasNames == true.
+/// \return enum_serialized_descriptor<payload_tags::unsigned_le, T>, i.e. the same on-wire
+///         unsigned integer representation as memcpy_serialized_descriptor, but additionally
+///         exposing T so that EnumTraits<T> can be looked up by downstream consumers.
+// This is false positive, Overload signatures are different.
+// coverity[autosar_cpp14_m3_2_3_violation : FALSE]
```
hoppe-and-dreams, line 1123: More coverity here

score/static_reflection_with_serialization/serialization/test/ut/test_serializer_visitor.cpp
```diff
@@ -971,4 +971,66 @@ TEST(clear_functionality_test, test_that_clear_function_can_clear_vector_of_int3
...
+
+TEST(serializer_visitor, named_enum_wire_format_unchanged)
+{
+    RecordProperty("ParentRequirement", "SCR-1633893");
+    RecordProperty("ASIL", "B");
+    RecordProperty("Description",
+                   "Check that registering EnumTraits<T> for an enum type only adds compile-time "
+                   "enumerator name metadata and does not alter its serialized wire representation.");
+    RecordProperty("TestingTechnique", "Requirements-based test");
+    RecordProperty("DerivationTechnique", "requirements-analysis");  // requirements
```
hoppe-and-dreams, line 1005: None of this makes sense in S-CORE, does it @4og ?

hoppe-and-dreams (commented):
Thanks for the contribution. Looks well in general. I added some minor findings to please address.

score/static_reflection_with_serialization/serialization/test/ut/test_serializer_visitor.cpp
4og: It actually does, but the values are slightly different in S-CORE. The best is to see them in this markdown: https://github.com/eclipse-score/baselibs/blob/main/.agents/skills/requirements-management-skill/test-to-requirement-linking.md

We currently have an epic to clean all of this up across the codebase: https://github.com/eclipse-score/baselibs/issues/522

For this PR, I would just make the `Description`, `TestingTechnique`, `DerivationTechnique` aligned with the S-CORE guidelines (basically, only the `TestingTechnique` needs to be turned to `("TestType", "requirements-based")`. For the rest I don't care for now.

score/static_reflection_with_serialization/serialization/include/serialization/visit_serialize.h
rmaddikery: Its fine for documentation of intent w.r.t to suppression until a new tool is introduced where the suppression could be removed OR carried over.

PR #533, opened 2026-08-31
coverage: adopt the LLVM coverage pipeline from score_tooling

### LINKED ISSUE #512: Integrate coverage infrastructure provided by score_tooling

dcalavrezo-qorix:
Closes #512

## What

Replaces the gcov/genhtml C++-only coverage flow with the shared LLVM source-based pipeline from `@score_tooling//coverage` (see the [adoption guide](https://github.com/eclipse-score/tooling/blob/main/coverage/README.md)). One `bazel coverage --config=llvm_cov` run now produces a single report for **C++ and Rust** with line + branch data, **exact 0% baselines** for untested in-scope files, a justification mechanism (`COV_JUSTIFIED` markers + `tools/coverage/coverage_justifications.yaml`, wired in from the start) and a threshold gate (**report-only**, `COVERAGE_THRESHOLD: "0"`, to be ratcheted once a baseline is agreed).

## Validation (local, x86_64-linux)

- 299/299 tests pass (5 QNX-only tests skipped; post-rebase on #535 and #471)
- **86.83% lines** (22706/26149), **79.94% branches** (4033/5045)
- 512 files in the report: 478 C/C++, 34 Rust; **42 files at exact 0%** (baseline mechanism verified, e.g. `containers_rust/lib.rs`, `stdout_logger/lib.rs`, vajson reader headers)
- Gate verified: exits non-zero at `COVERAGE_THRESHOLD=100`, zero at `0`
- `--lockfile_mode=error` clean; `//:format.check` and `//:copyright.check` green

## Scope completeness check (new)

`.github/tools/check_coverage_scope.sh` runs as the first step of the coverage workflow: it compares (1) every production `cc_library`/`rust_library` under `//score/...` against (2) everything transitively reachable from the scope and (3) the justified exclusions in `tools/coverage/scope_exclusions.txt`, failing on unlisted targets and on stale exclusions. Running it surfaced real gaps (orphaned tracing `interface_types` libs, unreachable `json:interface`, mock-only-consumed os wrappers) which are now in scope — the drop from the earlier 89.25% headline to 86.83% is those previously-missing files being counted, most at 0%.

## Scope

`tools/coverage:coverage_scope` lists the dependency-forest roots of **every non-testonly `cc_library`/`rust_library` under `//score/...`** — the same surface the legacy `bazel coverage //score/...` instrumented. The regeneration query is documented in the BUILD file. QNX-only libraries are excluded (not buildable on the Linux host; on-target coverage is tracked in eclipse-score/tooling#427). Select-branch implementation libs (e.g. `os:pthread_default`, nlohmann parser) are reached transitively through their public aliases; `vajson_parser` (non-default `base_library` branch) is listed explicitly with a visibility grant so its own unit tests count. Test scaffolding (`test_types`, `ffi_test_helpers`, mocks' stub default, `valgrind_on_host`) is out.

**Known gap (restricted path):** `score/mw/log/detail/utils/signal_handling` is a scope root with restricted visibility, but its BUILD file lives under the restricted path `score/mw/**` — the visibility grant must go through the internal-export process (cc @4og). Until then its files are absent from the report; every other `score/mw/log` root is in scope (their BUILD files needed no changes).

## Excluded tests (Clang-incompatibilities, documented in the workflow; all keep running in the GCC jobs)

| Test | Cause |
|---|---|
| `safecpp/aborts_upon_exception/...` | the library interposes `__cxa_allocate_exception`; GNU ld tolerates the duplicate against `libstdc++.a`, ld.lld rejects it (`duplicate symbol`). `toolchain_test` was already excluded (expected-fail, #7) |
| `safe_math/details:floating_point_environment_test` | relies on GCC FP-exception semantics; fails under plain Clang **without** any coverage flags (verified) |
| `os/linux/utils/test:network_interface_test` | file-scope `static StatMock` trips the gmock leak check only under `--dynamic_mode=off` static linking; test-hygiene follow-up candidate |

## Notes

- The existing LLVM 22.1.7 toolchain instance is extended (death-test coverage feature, `-lrt`/`-latomic`, `llvm_toolchain_llvm` for llvm-cxxfilt) — clang-tidy behavior unchanged (feature only active under `--config=llvm_cov`).
- Rust needs **no** custom toolchain: score_toolchains_rust ≥ 0.10.0 standard Ferrocene toolchains ship `llvm-cov`/`llvm-profdata`.
- The CI artifact name (`*_cpp_coverage_report`) is kept so `release.yml` keeps working; it now contains the unified C++ + Rust report (rename can follow separately).
- The cargo tarpaulin job is untouched (parallel-run period); Bazel coverage now reports Rust too.
- Visibility grants (`//tools/coverage:__pkg__`) remain on six orphan scope roots with restricted visibility: `json:interface`, `canary_wrapper`, `os/linux:linux`, `os/linux/utils:network_interface`, `memory:endianness`, `json/.../vajson_parser` — kept for now per review discussion (#535/#536 retired the other candidates: cpuid, make_dev, file_stream, netdb, select). Deprecated shims (`memory/shared:atomic_interface`, `memory:string_literal`) are excluded from the scope per review; the deprecated-target warnings the scope surfaces (e.g. `datetime_converter`) are pre-existing.


score/os/BUILD
```diff
@@ -28,6 +28,8 @@ cc_library(
     features = COMPILER_WARNING_FEATURES,
     tags = ["FFI"],
     visibility = [
+        # Coverage scope root (//tools/coverage, eclipse-score/baselibs#512).
+        "//tools/coverage:__pkg__",
```
4og: Doesn't seem to be needed.

Same applies to other targets below

score/os/BUILD
```diff
@@ -340,6 +348,8 @@ cc_library(
     features = COMPILER_WARNING_FEATURES,
     visibility = [
         ":__subpackages__",
+        # Coverage scope root (//tools/coverage, eclipse-score/baselibs#512).
```
4og: Won't be needed hopefully. See #535

score/os/BUILD
```diff
@@ -369,6 +379,8 @@ cc_library(
     hdrs = ["cpuid.h"],
     features = COMPILER_WARNING_FEATURES,
     visibility = [
+        # Coverage scope root (//tools/coverage, eclipse-score/baselibs#512).
```
4og: Addressed by #535

score/os/BUILD
```diff
@@ -407,6 +419,8 @@ cc_library(
     hdrs = ["makedev.h"],
     features = COMPILER_WARNING_FEATURES,
     visibility = [
+        # Coverage scope root (//tools/coverage, eclipse-score/baselibs#512).
```
4og: Addressed by #535

score/os/BUILD
```diff
@@ -421,6 +435,8 @@ cc_library(
     hdrs = ["file_stream.h"],
     features = COMPILER_WARNING_FEATURES,
     visibility = [
+        # Coverage scope root (//tools/coverage, eclipse-score/baselibs#512).
```
4og: Addressed by #535

tools/coverage/BUILD
```diff
@@ -0,0 +1,129 @@
...
+# roots of the dependency forest are listed - the scope aspect walks
+# transitive in-workspace deps. QNX-only libraries are excluded (they cannot
+# be built on the Linux host; on-target coverage: eclipse-score/tooling#427).
+#
+# Regenerate the list after adding/removing libraries:
+#   bazel query 'kind("cc_library|rust_library", //score/...)
+#       except attr(testonly, 1, //score/...)
+#       except attr("target_compatible_with", "qnx", //score/...)'
+# then keep only targets no other target in the result depends on.
+score_coverage_scope(
```
4og, line 28: We'll have a really hard time to maintain this list :(

score/memory/shared/BUILD
```diff
@@ -32,6 +32,8 @@ cc_library(
     features = COMPILER_WARNING_FEATURES,
     tags = ["FFI"],
     visibility = [
+        # Coverage scope root (//tools/coverage, eclipse-score/baselibs#512).
```
4og: I wouldn't measure coverage for this deprecated target and use `@score_baselibs//score/concurrency:atomic_interface` instead

score/memory/BUILD
```diff
@@ -18,6 +18,8 @@ load("@score_baselibs//score/quality/clang_tidy:extra_checks.bzl", "clang_tidy_e
 load("@score_docs_as_code//:docs.bzl", "docs_bundle")
 
 package(default_visibility = [
+    # Coverage scope root (//tools/coverage, eclipse-score/baselibs#512).
```
4og: I'm not sure why it's needed

score/os/BUILD
dcalavrezo-qorix: Right — those four were collateral from a bulk edit and not scope roots; removed in 15fefdd45 (now part of the rebase).

score/os/BUILD
dcalavrezo-qorix: Rebased on main with #535 — netdb is public now, grant dropped (7acb8f60e).

score/os/BUILD
dcalavrezo-qorix: Gone with #535 — target retired; removed from the coverage scope in 7acb8f60e.

score/os/BUILD
dcalavrezo-qorix: Gone with #535 — target retired; removed from the coverage scope in 7acb8f60e.

score/os/BUILD
dcalavrezo-qorix: Gone with #535 — target retired; removed from the coverage scope in 7acb8f60e.

score/memory/shared/BUILD
dcalavrezo-qorix: Agreed — dropped the deprecated shim from the scope and reverted the grant in 7acb8f60e; its replacement (//score/concurrency:atomic_interface) is already covered transitively. Same for //score/memory:string_literal.

score/memory/BUILD
dcalavrezo-qorix: It was there for //score/memory:endianness — the only dependency-forest root in the package relying on the package default. Replaced the package-default change with an explicit visibility on endianness only, with a comment (7acb8f60e).

tools/coverage/BUILD
dcalavrezo-qorix: Understood — some context: the list is generated, not curated (the query is in the header comment), and listing only dependency-forest roots keeps it ~90 entries instead of ~300. Maintenance burden in practice: a NEW library is only missed if nothing else depends on it AND it has no tests (then it silently misses its 0% row — same blind spot the old gcov flow had for every untested file). Options to harden it: (a) a small CI check that re-runs the query and diffs against the BUILD list, (b) periodic regeneration, or (c) accepting drift and reviewing the list at release points. Happy to add (a) as a follow-up if you want it.

4og: I vote for a small CI check

paulquiring: I, too, think the CI check is the best option.

4og (approved):

PR #539, opened 2026-09-02
Remove Linux constraint from require_typed_memory

devendrapat:
Removed contradictory Linux constraint from require_typed_memory config_setting as it is blocking score_communication from using this config_setting, as this was blocking score_communication from using the config_setting in their code. In this PR, changes were made to use a bool_flag constraint to set require_typed_memory flag so that score_communication can depend on it without issues.

Please refer https://github.com/eclipse-score/communication/pull/1022 for the error faced by score_communication while using require_typed_memory_config_setting.


4og:
@hoppe-and-dreams please have a look

score/analysis/tracing/generic_trace_library/flags/BUILD
```diff
@@ -15,7 +15,6 @@ config_setting(
     name = "require_typed_memory",  #This is a stub and should always be false
     constraint_values = [
         "@platforms//os:qnx",
-        "@platforms//os:linux",  # Contradicts qnx — ensures this config_setting never matches (stub)
```
hoppe-and-dreams, line 18: If we do this, all S-CORE LoLa builds on QNX would go to typed memory. I do not think this is the intended (and aligned) consequence.

If it is indeed the desired outcome, at least the comment 3 lines above must be changed to match what the code does.

hoppe-and-dreams (changes requested):

score/analysis/tracing/generic_trace_library/flags/BUILD
devendrapat: Hello @hoppe-and-dreams,
Changes were made to use a bool_flag constraint to set require_typed_memory flag so that score_communication can depend on it without issues. Also the comment above was modified a bit to convey that the flag is not used anywhere in the repository, and the bool_flag is False by default.

hoppe-and-dreams (approved):

4og (approved):

PR #540, opened 2026-09-02
AGENTS.md: clarify copyright header scope and docs layout

4og:
Copyright headers are only enforced for a specific set of file extensions, not every new file. This clarifies the copyright section accordingly, and documents that component docs are split between `score/<component>/docs/` and `docs/baselibs/components/<component>/docs/` depending on migration status.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -9,10 +9,11 @@
 ```
 score/          # C++ AND Rust libraries side by side (concurrency, containers, containers_rust, filesystem,
                 # json, memory, network, os, result, allocator, pal, sync, thread, testing_macros, log_rust, etc.)
+                # Most components have a docs/ subfolder with requirements, architecture, detailed design, and safety analysis.
 src/            # Deprecated Bazel alias stubs. No real sources here.
 examples/       # Integration and usage examples (C++ and Rust)
 third_party/    # External dependency BUILD files (acl, openssl, libcap2, etc.)
-docs/           # Sphinx documentation sources
+docs/           # Feature-level and module-level Sphinx docs.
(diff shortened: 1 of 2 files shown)


AGENTS.md
```diff
@@ -77,14 +78,15 @@ This runs all format checks (C++, Python, Rust, Starlark, YAML) as test targets.
 
 ### Copyright headers
 
-Copyright headers are enforced in this repository; CI will fail if they are missing or incorrect. If unsure how the header should look for a given file type, don't guess: run `copyright.fix` and let it generate the header for you.
+Copyright headers are enforced for specific file extensions (`bazel`, `BUILD`, `bzl`, `c`, `cpp`, `h`, `hpp`, `ini`, `py`, `rs`, `rst`, `sh`, `yaml`, `yml`); CI will fail if a covered file's header is missing or incorrect.
```
fbaeuerle, line 81: the complete list is specified here:
https://github.com/eclipse-score/tooling/blob/main/cr_checker/resources/templates.ini

fbaeuerle (approved):
the list of file extensions is not complete and is easily outdated. you may consider adding a reference to the template config, but i cannot judge whether that helps or not.

AGENTS.md
4og: Actually, we use only a subset:
https://github.com/eclipse-score/baselibs/blob/6932bd28c6e66789bdf3ef4cf662183ed0a9791e/BUILD#L135-L149

I don't know what's better, force the agent to tinker with the BUILD file (and put also irrelevant stuff into the context) or just accept duplication.

PR #562, opened 2026-09-08
fix(log_rust): replace unstable NonNull::from_ref with manual cast

ShoroukRamzy:
- Remove compile crashes on the safety-certified Ferrocene Rust compiler.

- **Removes the patch** used here https://github.com/eclipse-score/inc_security_crypto/pull/204 

- **Unblock** https://github.com/eclipse-score/inc_security_crypto/issues/173


pawelrutkaq (changes requested):
wrong change, from_ref is stable

ShoroukRamzy:
> wrong change, from_ref is stable

Hi @pawelrutkaq, you are right, score crypto fall to older version of rustc compiler 1.76 coming from toolchain rust, that's why baselibs failed there in the host execution phase, i updated the code so it can cleanly build in both the newer rust compiler in baselibs standalone repo and the host older compiler.

pawelrutkaq:
there shall not be such a case. We support 1.91 + and thats it and the host and target compiler is feroceen.

PR #572, opened 2026-09-10
utils: replace legacy control-flow-analysis/Verifies tags with comp_req RecordProperty links

Nikhil2206:
## Summary
Part of #522. Replaces the legacy `control-flow-analysis`/`Verifies` `RecordProperty` tags in `base64_test.cpp` with the current `TestType`/`DerivationTechnique`/`PartiallyVerifies: comp_req__utils__base64_encoding`/`comp_req__utils__base64_decoding` linkage convention.

Validated: `bazel test --config=bl-x86_64-linux //score/utils:base64_unit_test` → PASSED.

## Deliberately left unlinked / flagged issues
None — all 8 tests in this file already had a 1:1 mapping to the base64 encoding/decoding requirements; this is a pure metadata-convention modernization, not a new-linkage pass.

### DIFF
--- a/score/utils/base64_test.cpp
+++ b/score/utils/base64_test.cpp
@@ -21,8 +21,9 @@ namespace score::utils::test
 
 TEST(Base64Test, EncodeEmptyString)
 {
-    ::testing::Test::RecordProperty("TestType", "control-flow-analysis");  // control flow
-    ::testing::Test::RecordProperty("Verifies", "::score::utils::EncodeBase64()");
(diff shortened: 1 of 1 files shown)


4og:
This PR is generally ok, but it needs a follow up to address the missing traceability for https://github.com/eclipse-score/baselibs/blob/v0.2.13/score/utils/meyer_singleton and https://github.com/eclipse-score/baselibs/blob/v0.2.13/score/utils/static_destruction_guard.h

4og (approved):

PR #574, opened 2026-09-10
json: update the documentation regarding the usage of GetAttribute

### LINKED ISSUE #487: Provide optional-field accessor helpers in score::json (GetAttribute usability)
### What

When parsing JSON configuration with `score::json`, consumers cannot cleanly express the common pattern "use the default if the key is missing, fail fast if the key exists but has the wrong type", because:

1. `GetAttribute<T>(obj, key)` does not accept a raw `const Object&` (only `reference_wrapper<const Object>` or `Result<...>`), forcing callers to wrap with `std::cref()` or fall back to manual `find -> second.As<T>()`.
2. `Result::value_or(default)` swallows type errors together with "key not found", so a mistyped field silently falls back to the default instead of producing a clear error.

While implementing the TimeSlave JSON config parser (https://github.com/eclipse-score/time/pull/158) I had to duplicate the same ~10-line helper wrapper. Other components parsing JSON configs will likely hit the same issue.

### Acceptance Criteria (DoD)

The score::json documentation (README or header docs) is updated to include:
An example showing GetAttribute<T>(std::cref(obj), key) as the standard way to access object fields, clarifying that std::cref/std::ref is the intended calling convention.
An example of the "default on missing key, fail fast on type error" pattern, demonstrating how to distinguish Error::kKeyNotFound from other errors without using value_or.

### How

_No response_

mihajlo-k:
Closes #487

Documentation-only change, per the conclusion in #487 that the issue is discoverability, not a missing API.

- `README.md`: documents `GetAttribute<T>(std::cref(obj), key)` as the intended calling convention and why there is no `const Object&` overload; adds the "default on missing key, fail on wrong type" pattern via an explicit `kKeyNotFound` check.
- `object.h`: short doc comment stating the same convention at the point of use.

Also fixes the existing nested-attribute example, which did not compile, and a dead link to `getters_test.cpp`.

### DIFF
--- a/score/json/README.md
+++ b/score/json/README.md
@@ -9,6 +9,9 @@
     - [Accessing numbers from a JSON document](#accessing-numbers-from-a-json-document)
     - [Accessing strings from a JSON document](#accessing-strings-from-a-json-document)
     - [Simplified usage](#simplified-usage)
+      - [Calling convention: pass the object by `std::cref`](#calling-convention-pass-the-object-by-stdcref)
+      - [Accessing a nested attribute](#accessing-a-nested-attribute)
+      - [Optional attributes: default on missing key, fail on wrong type](#optional-attributes-default-on-missing-key-fail-on-wrong-type)
     - [Declarative parsing of JSON data](#declarative-parsing-of-json-data)
   - [Design](#design)
   - [Constraints](#constraints)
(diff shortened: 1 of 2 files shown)


score/json/README.md
```diff
@@ -224,16 +256,44 @@ similar to:
...
+    const auto geometry = GetAttribute<json::Object>(widget, "geometry");
+    const auto size = GetAttribute<json::Object>(geometry, "size");
+    return GetAttribute<std::uint64_t>(size, "width");
 }
 ```
 
-See the [unit tests](./getters_test.cpp) for details.
+Only the first call needs `std::cref`. Every intermediate `Result` is forwarded
+directly, and an error of any step is propagated to the final result, so the
+error only needs to be checked once at the end.
```
gordon9901, line 270: This is a great clarification — the fact that intermediate Results propagate errors automatically (so only a single end-of-chain check is needed) is exactly the detail that wasn't obvious before.

gordon9901:
Thanks for the quick turnaround — this documentation update fully addresses the discoverability issue raised in #487.
Approved from my side, ready to merge.

gordon9901 (approved):

4og (approved):

PR #576, opened 2026-09-10
ci: consolidate linting with rules_lint

fbaeuerle:
fail_on_violation=true made Bazel drop the aspect's outputs whenever
clang-tidy actually found something, since sandboxed actions lose all
declared outputs on a non-zero exit. Build with fail_on_violation=false
instead and fail the job at the end, once reports are already collected.


.github/workflows/lint.yml
```diff
@@ -37,21 +41,42 @@ jobs:
         uses: eclipse-score/cicd-actions/setup-bazel-cache@659dcbed63f6b7fbde88c7850125ea0a0a92f939
         with:
           unique-cache-name: ${{ github.workflow }}-${{ github.job }}
-      - name: Bazel Build with clang-tidy
+      - name: Bazel build with clang-tidy
```
4og: Is it sensible to join three separate bazel build steps into one and also combine the three configs in `.bazelrc` into a single `lint` config?

.github/workflows/lint.yml
```diff
@@ -60,54 +85,61 @@ jobs:
           npm install --no-save --prefix /tmp/sarif-multitool @microsoft/sarif-multitool@5.7.0
           chmod +x /tmp/sarif-multitool/node_modules/@microsoft/sarif-multitool-linux/Sarif.Multitool
           echo "/tmp/sarif-multitool/node_modules/.bin" >> "$GITHUB_PATH"
-      - name: Clean clang-tidy SARIF reports
+      - name: Build merged SARIF per linter
```
4og: Having separate SARIFs per tool looks too inflexible to me. If we decide to add an extra linter, then it will require changes in a lot of places. Plus it introduces duplication in workflow files.

Doesn't `sarif-multitool` support merging sarifs produced by multiple tools? I see the tool.driver is filled in individual sarifs, so I assume the multitool should be able to take this into account.

Relevant GitHub docs: https://docs.github.com/en/code-security/reference/code-scanning/sarif-files/sarif-support#sariflog-object

4og (changes requested):

4og (approved):

PR #585, opened 2026-09-15
Move score tooling with dev_dependency = TRUE to to tools

antonkri:
This pull request restructures Bazel development tooling by moving tool targets (copyright, formatting, license checks, and compile commands generation) out of the root `BUILD` file into a new `tools/BUILD` file. This change helps keep the root build configuration free from dev dependencies and clarifies the separation between core build logic and developer tooling. It also updates documentation, pre-commit hooks, and CI workflows to reference the new tool locations.


tools/BUILD
```diff
@@ -0,0 +1,80 @@
...
+# information regarding copyright ownership.
+#
+# This program and the accompanying materials are made available under the
+# terms of the Apache License Version 2.0 which is available at
+# https://www.apache.org/licenses/LICENSE-2.0
+#
+# SPDX-License-Identifier: Apache-2.0
+# *******************************************************************************
+
+# CI tooling entrypoints backed by @score_tooling and @hedron_compile_commands.
```
4og: Comment is wrong. It's not only about CI

4og (approved):

